PR #2, opened 2020-03-16
feat: Adding org module to example

rjerrems:
The purpose of this step is to setup top level shared folders, monitoring projects, org level logging & set baseline security settings through organizational policy


ericyz:
@rjerrems - Do you think if it's necessary to draft a sample cloudbuild-plan & cloudbuild-apply file as a template? I could help if you think it brings value

rjerrems:
Good point @ericyz - I have added cloud build configuration files.

rjerrems:
It looks like @ericyz already approved this one, so should be ready for final review from you now @morgante

PR #3, opened 2020-03-19
feat: Add initial networking code

rjerrems:
The purpose of this step is to setup shared VPCs with default DNS, NAT, Private Service networking and baseline firewall rules.

### DIFF
--- a/2-networks/cloudbuild-tf-plan.yaml
+++ b/2-networks/cloudbuild-tf-plan.yaml
@@ -1,0 +1,41 @@
+# Copyright 2020 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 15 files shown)


2-networks/cloudbuild-tf-plan.yaml
```diff
@@ -0,0 +1,41 @@
...
+# limitations under the License.
+
+timeout: 1200s
+steps:
+- name: gcr.io/$PROJECT_ID/terraform
+  entrypoint: /bin/bash
+  args:
+  - -c
+  - |
+    gcloud config set auth/impersonate_service_account ${_TF_SA_EMAIL}
```
ericyz: I believe that it would be authenticated as cloudbuild service account from the next step onwards. The workspace is mounted to a new container for each step.

rjerrems: From my testing, it looks like the subsequent steps still use the same configured impersonation. This could be because the gcloud config is mounted in the workspace, but could be worth digging into more in case this fails in where the container might run in an alternate location without preserving the configured state of gcloud - do you think you could help looking into that?

ericyz: Interesting one. I tested as well and it seems using the cloud build service account. Do u mind sharing your experience how you tested it?

ericyz: Interesting one. I tested as well and it seems using the cloud build service account. Do u mind sharing your experience how you tested it?

Issue #12, opened 2020-04-23
Make project example folder structure and file naming conventions clearer


rjerrems:
For the projects repo, it should be clear that the top level folder is a business unit or team. We should rename this to something like:

example_product or example_team etc

Following on from this, we should also use that in the naming convention for the tf files. For example

example_product_sample_project.tf

@ericyz - what do you think?

rjerrems:
to expand on this and comment in https://github.com/terraform-google-modules/terraform-example-foundation/pull/14

the file naming convention could be something like

<business_unit>_<product/app>_sample_projects.tf

rjerrems:
Fixed in #14

PR #14, opened 2020-04-24
update top-level folder name

ericyz:


rjerrems:
This is to address this issue https://github.com/terraform-google-modules/terraform-example-foundation/issues/12

rjerrems:
I think if this is a generic top level folder, like products - it should probably move to the 1-org step. I would however argue that we should avoid adding something at the top level unless it is adding value given the hierarchy only goes 4 layers deep. What I would advocate is a hierarchy like this:

- business unit / product area (folder)
-- product/app (folder)
--- prod & nonprod (folders)
---- product widget service (projects)
---- product foo service (projects)

That way each business unit can have N products/apps and each app can have N services in the prod/non prod folders. It also gives 3 layers which are practical to delegate IAM if a user wishes, whereas a top level folder like `products` or `app` doesn't really provide value other than being a convenient container visually.

What do you think?

morgante:
We should choose one of the hierarchies from go/org-hierarchy-options

3-projects/folders.tf
```diff
@@ -14,7 +14,19 @@
...
+ *****************************************/
+resource "google_folder" "business_unit" {
+  display_name = "example-business-unit"
+  parent       = "organizations/${var.org_id}"
+}
+
+/******************************************
+  Team Folders
+ *****************************************/
+module "standard_project_folders" {
```
rjerrems, line 28: for consistency, it probably makes sense to call this something like `example_team_folders`

3-projects/folders.tf
```diff
@@ -14,7 +14,19 @@
  * limitations under the License.
  */
 
-resource "google_folder" "app" {
-  parent       = "organizations/${var.organization_id}"
-  display_name = "app"
+/******************************************
+  Business Unit Folders
+ *****************************************/
+resource "google_folder" "business_unit" {
```
rjerrems, line 20: nit: to help clearly label what this is would be good to have something like `example_business_unit` for resource name

PR #24, opened 2020-04-29
refactor: Harden network & add default example subnets

rjerrems:
This change:
- Adds default example subnets, with secondary ranges
- Deletes default egress routes
- Adds tag based egress route for internet access
- Sets up private access default route via IGW
- Adds DNS for private gcr.io and googleapis.com
- Refactors submodule structure and adds some variables for convenience


rjerrems:
@morgante - when you get a chance, would be great to get your review on this one as it is changing a few things and setting some defaults I want to make sure you are onboard with.

2-networks/main.tf
```diff
@@ -40,15 +40,59 @@ module "shared_vpc_nonprod" {
   project_id           = local.nonprod_host_project_id
   default_region       = var.default_region
   network_name         = "shared-vpc-nonprod"
-  private_service_cidr = "10.1.0.0/16"
+  private_service_cidr = "10.200.0.0/22"
```
morgante, line 43: What's the relationship between the private service CIDR and the subnets?

2-networks/modules/standard_shared_vpc/dns.tf
```diff
@@ -0,0 +1,112 @@
...
+  networks {
+    network_url = module.main.network_self_link
+  }
+}
+
+/******************************************
+  Private Google APIs DNS Zone & records.
+ *****************************************/
+
+resource "google_dns_managed_zone" "private_googleapis" {
```
morgante, line 36: Could we use the DNS module to handle this? https://github.com/terraform-google-modules/terraform-google-cloud-dns

2-networks/modules/standard_shared_vpc/firewall.tf
```diff
@@ -0,0 +1,70 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/******************************************
+  Default firewall rules
+ *****************************************/
+
+// Allow SSH when using the allow-ssh tag for Linux workloads.
```
morgante, line 21: We should allow these default rules to be disabled.

Also I'd make it clear this is allowing SSH from IAP. Maybe call it `allow-iap-ssh`.

2-networks/modules/standard_shared_vpc/main.tf
```diff
@@ -19,41 +19,39 @@
...
 
   subnets          = var.subnets
   secondary_ranges = var.secondary_ranges
 
   routes = [
     {
-      name              = "gcp-windows-activation"
-      description       = "Route through IGW to allow Windows kms activation."
+      name              = "egress-internet"
+      description       = "Tag based route through IGW to access internet"
```
morgante, line 35: I think the default pattern should probably be Cloud NAT instead?

rjerrems: So this is still using Cloud NAT, its just by default you do not have a path to the internet if that makes sense?

2-networks/modules/standard_shared_vpc/firewall.tf
rjerrems: Yeah that makes sense, I will update.

2-networks/modules/standard_shared_vpc/dns.tf
rjerrems: Sure thing, this will make it more concise

2-networks/main.tf
rjerrems: So the private service CIDR is the range which gets allocated for Cloud SQL and other services which use that methodology. When you for example launch your first private MySQL instance, it will carve a /24 out of that allocated block automatically. If you want that range to be routable to on-prem or other networks, it also needs to be on a non-overlapping IP range with other subnets.

rjerrems:
@morgante - thanks for that feedback, I have pushed some updates to address now.

2-networks/modules/standard_shared_vpc/main.tf
rjerrems: For context, the PR changes the default behavior to:
- Access and routes available to Google APIs but no internet access
- egress-internet tag must be applied to reach anything on the internet via Cloud NAT

2-networks/main.tf
rjerrems: https://cloud.google.com/vpc/docs/configure-private-services-access

2-networks/modules/standard_shared_vpc/main.tf
morgante: Got it.

PR #37, opened 2020-05-19
initial commit for app iac pipeline

ericyz:


morgante:
I'm not sure we should put particular app pipelines in this repo? The rest of the repo can mostly be adopted as-is, but individual apps are so different that I think a generic implementation is less useful and probably belongs elsewhere.

ericyz:
I think I have named it in a confusing way. It's to create application iac repo and cloud build trigger for app team. It can be used as a common starting point for them as a native GCP CI/CD solution.

rjerrems:
Yeah I think there is some value having something like this, but probably need to it optional rather than the default as many customers will have existing tooling they want to use. (e.g an optional sub-module you can use in the projects pipeline, which sets everything you need up to run dedicated app specific cloud build pipelines with the right permissions over the projects you have created).

What I think would be most useful right now is an example application / infra deployment, so people can see something practical they can do with the projects that get created. What do you think @morgante ?

morgante:
> What I think would be most useful right now is an example application / infra deployment, so people can see something practical they can do with the projects that get created. What do you think @morgante ?

Yes, I agree that would be useful. I'd like to land it in a different (new) repo though, something like `terraform-example-application`. Does that sound reasonable?

rjerrems:
Yeah that sounds good to me, I think we can link to it in the docs from here and make sure that it works nicely with the example foundation that gets created here. Lets break this change out to provide an optional submodule in `3-projects` to configure cloud build based pipelines dedicated for your project/s and create a seperate repo `terraform-example-application` to demonstrate deploying an app. (which does not depend on that optional submodule)

PR #39, opened 2020-05-26
Langley stage

langleyce:
Includes new environment structure for Production. Once approved and naming convention is final, I can update the remaining envs.


morgante:
I'm not sure why there's a separate `Production` directory?

rjerrems:
Thanks for the contribution! It does seems like we are introducing a bunch of environments & networks which will complicate the deployment / getting started for users of the code. Our general approach so far has to try and make core base foundation lean and simple, allowing users to choose to add additional components through submodules or modifications of the codebase. It would be good to maintain consistency with that approach if we can. (i.e I am not sure everybody will need the environments supplied in this change, but being able to easily stamp out new environments is useful) 

Also we did intentionally structure the folders with environments lower in the hierarchy, so would be good to cover why changing that will improve the structure?

To help us understand  @mikelaramie / @langleyce - do you think you could update the description to provide some more detail of what is in the box / what your trying to achieve to help with review and feedback?

PR #40, opened 2020-05-27
1-org Separate IAM resources and add monitoring workspace permissions

davidholsgrove:
Separate out the IAM resources for clarity, and add a IAM Member with `roles/monitoring.editor` to the Monitoring Projects


rjerrems:
Thanks for the contribution @davidholsgrove - can you please run `make docker_generate_docs` and fix the output of `make docker_test_lint` it looks like some of the terraform needs a `terraform fmt`

PR #45, opened 2020-06-03
Seed folder creation, dev_folder variable

mikelaramie:
This code adds logic to put the seed projects in a folder (named "seed"  by default).

Additionally introduced the "dev_folder" construct that allows these scripts to be run against a folder instead of an organization.  If the dev_folder variable is set, it will use that folder as the root; otherwise it will use the organization as the root.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,6 +18,11 @@ crash.log
 # version control.
 #
 # example.tfvars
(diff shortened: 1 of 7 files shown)


mikelaramie:
Closing this; just realized I need to add the logic to the other modules and then I'll submit a new PR

PR #46, opened 2020-06-03
Added dev_folder construct

mikelaramie:
Round 2 - sorry for the confusion.  This time, I've added the dev_folder construct to all modules.


1-org/folders.tf
```diff
@@ -20,7 +20,7 @@
 
 resource "google_folder" "common" {
   display_name = "common"
-  parent       = "organizations/${var.org_id}"
+  parent       = var.dev_folder != "" ? "folders/${var.dev_folder}" : "organizations/${var.org_id}"
```
morgante, line 23: One suggestion: instead of putting this logic throughout the codebase I'd suggest consolidating it in a local:

```
locals {
  parent = var.dev_folder != "" ? "folders/${var.dev_folder}" : "organizations/${var.org_id}"
```

rjerrems: +1 this will make it a lot easier to refactor

mikelaramie:
Makes sense; I'll refactor it.  Thanks for the feedback!

0-bootstrap/terraform.tfvars
```diff
@@ -0,0 +1,27 @@
+/**
```
morgante, line 1: Looks like this was accidentally committed?

mikelaramie: Spot on.  Let me remove that.  I was going to submit a future PR modifying the .gitignore file a bit and haven't done that yet.

PR #48, opened 2020-06-05
update bootstrap readme of IAM premissions

ericyz:


ericyz:
This is to address the permission required to create folder in the 0-bootstrap regarding with recent change

quinnturner:
@ericyz random person here, but consider using `1.` instead of `5.` for the ordering since Markdown auto-orders.

ericyz:
Thanks @quinnturner ! Good pick up! Fixed

rjerrems:
It might be best to link to the permissions section of that README instead @ericyz https://github.com/terraform-google-modules/terraform-google-bootstrap#permissions otherwise it LGTM

Issue #47, opened 2020-06-03
Change folder hierarchy to split environments


morgante:
Currently, we use a "flat" hierarchy in each component where all resources for a component are contained in a single Terraform state. This increases the blast radius and makes it challenging to test changes in nonprod before productionizing.

I'd like to split this repo to have folders for each environment (under the components), like the structure described [here](https://docs.google.com/document/d/1lpuK7salw2gUPdpQByPFAS4IvmM2aeHE5eFx5W_WTd8/edit#heading=h.wkpi8bw1p19k):
```
├── foundations
│   ├── folders
│   │   └── shared
│   │       └── folders.tf
│   ├── logging
│   │   ├── nonprod
│   │   │   └── export.tf
│   │   └── prod
│   │       └── export.tf
│   ├── network
│   │   ├── nonprod
│   │   │   └── network.tf
│   │   └── prod
│   │       └── network.tf
│   ├── org-iam
│   │   └── shared
│   │       └── iam.tf
│   ├── org-policies
│   │   └── shared
│   │       └── org-policies.tf
│   ├── Jenkinsfile
│   ├── backend.tf
│   ├── common-providers.tf
│   ├── common-variables.tf
│   ├── main.tf
│   └── output.tf
```

morgante:
@rjerrems @ericyz What are your thoughts here?

rjerrems:
Yeah I agree with this in principle, we may also need to track some work on the bootstrap repo to change the way the triggers work to something more path based - or support that as an option? A couple of questions:

- Are you also proposing we removing the current defined numbered layers 1-org, 2-networks etc?
- Will the modules for each of those ^ still live in those layers or be in a central location?

morgante:
> Yeah I agree with this in principle, we may also need to track some work on the bootstrap repo to change the way the triggers work to something more path based - or support that as an option?

Yes, I think we'll have to do some work there.

> Are you also proposing we removing the current defined numbered layers 1-org, 2-networks etc?

No, we should still keep the top-level layers but just add the environment folders underneath them

> Will the modules for each of those ^ still live in those layers or be in a central location?

For now, I think we should just add them into this repo and assume they'll be pushed to a central location.

rjerrems:
Yeah okay cool, I am onboard if we keep the same top level folder structure -  it is a good idea to make it easier to roll out changes between environments, although we may still have some things that cut across the organization.

xingao267:
[...]

PR #49, opened 2020-06-05
Attempting to break out the Org repo structure

mikelaramie:
Hey @morgante and @rjerrems - hoping you can take a look at this.  I have two concerns over the way I implemented this:

1) Calling so many project factory and org policy modules repeated generated a .terraform/ directory of around 5.1GB when I ran it
2) I'm not entirely sure I'm doing this in the most efficient fashion

Any changes/suggestions?


0-bootstrap/terraform.tfvars
```diff
@@ -0,0 +1,27 @@
+/**
```
morgante, line 1: We shouldn't commit this.

mikelaramie:
The idea was to restructure "1-org" to align with the "Foundations" repo laid out in our Standards doc. If we think that it's better to just provide the individual logic pieces up in a top folder, I'm open to the discussion; but I'm not on the same page yet.  

I'm also open to the idea that I'm completely misunderstanding something here.

morgante:
Got it. I think in this case we should think of this entire repo as matching the `foundations` repo laid out in the standards doc. So you're going in the right direction, but we should move things up one level (to the root).

There's also no need to have a main.tf which composes the different components. The components can be individually invoked via Cloud Build.

We do have the projects collapsed in here too, but a customer could optionally split that out into a separate repo.

mikelaramie:
The way I was interpreting Rohan's work on this particular repo in its current state was:

- 0-bootstrap - this is a standalone workflow that stands up a seed project, a cloud build project, and 3 different CSRs with build triggers for 1-org ("gcp-org"), 2-networks ("gcp-networks") , and 3-projects ("gcp-projects")

- 1-org - this is intended to align to the "Repo 1: Foundations" repo from the Standards doc, and currently is supposed to get checked in to the "gcp-org" repo built in 0-bootstrap

- 2-networks - This manages shared VPCs to be used in the example foundation, and currently is supposed to get checked into the "gcp-networks" repo built in 0-bootstrap.  Work is scheduled to align this to the Enterprise Foundations design, but hasn't been started yet (per our comment convo in the Standards doc)

- 3-projects - This is intended to align to the "Repo 2: Project Creation" repo from the Standards doc (although that work hasn't been done yet),  and is currently supposed to get checked into the "gcp-projects" repo built in 0-bootstrap.  

As part of the deployment, it's up to the customer currently to split these 4 repos out from copying this repo.  Do you think it makes more sense to split them here, or keep them together?

morgante:
Realistically, I think that's probably overly complicated.

It leaves the bootstrap config stranded, while not making this repo easily forkable. Instead, I think we should simplify and treat this repo as the main "monorepo" for configuring foundations. It mostly matches "gcp-repo" (`1-org`) but also has examples of projects checked in.

Customers could then optionally split components out into separate repos, but I don't think we need to set that up by default. When you're first getting started, a separate repo for networking might not be needed (for example).

@rjerrems What do you think?

rjerrems:
Yeah I agree that this should remain the mono repo that can be easily forked, we have already had several people use it directly in this context without necessarily using the cloud build additions provided by default. (either run manually by a user or in a 3P CI/CD tool)

I think perhaps we are just getting our wires crossed with the structure provided here:

https://github.com/terraform-google-modules/terraform-example-foundation/issues/47

The current changes add some obfuscation in terms of what terraform is doing e.g org-1/networks creates projects not networks. We also have not really reduced the blast radius as all of the resources are created in the same terraform run & state backend. A few things I think we need to agree on are:

1. What does the rough directory layout look like and how does that map to different terraform state backends?
2. How will these be triggered / run? Without having a solution for this, we are hurting the usability of the repo.
3. What are the baseline environments? Seems we are changing from nonprod to dev here.

We need to be careful to not make it too complicated for someone to pick up and run, either themselves or through cloud build. Whilst this does not map exactly to the standards doc, a simpler starting point could be something like this:

```
org-1/
├── nonprod
│   ├── backend.tf
│   ├── main.tf
│   ├── outputs.tf
│   └── variables.tf
├── prod
│   ├── backend.tf
│   ├── main.tf
│   ├── outputs.tf
│   └── variables.tf
└── shared
    ├── backend.tf
    ├── main.tf
    ├── outputs.tf
    └── variables.tf
```

Although separating into different environments, whilst providing a mechanism for lower blast radius will make it inherently more complicated to run end to end.

mikelaramie:
I want to avoid deviating from [#47 ](https://github.com/terraform-google-modules/terraform-example-foundation/issues/47) /Standards doc here, because we're using that as the foundation for the repository in the Security Best Practices whitepaper as well.  Deviating from that here gets us back into a conversation of keeping this as two different repos, which we wanted to avoid.  My primary goal is to have an example repo following our Standards doc for the Security Best Practices guide.

- **dev vs. nonprod** - I'm more than happy to switch it back to nonprod
- **networks** - We could move the networks logic into there as well and remove the separate Networks root folder/repo structure
- **projects** - We appear to recommend splitting out Projects and Foundations in our Standards; as well as keeping separate states/backends per business unit, per environment.  It seems to me that that should be a separate repo since it's going to have different build requirements in either Jenkins or Cloud Build,
- **triggers/runs** - I'm more than happy to provide the yaml I'm using to test it in Cloud Build if we think there's benefit there, but our Standards and the SBP paper both reference Jenkins.  We could theoretically provide examples for both.

rjerrems:
Yeah I think there is still some confusion, at least on my behalf on the path forward - will talk tomorrow and figure out a solution

rjerrems:
Ok to get this moving, lets do the following:

- rebase against develop branch and change the PR to merge into develop (to keep master stable whilst we are refactoring)
- Make this change focused only on shared org wide related concerns, with environment specific things moving into the environment module as outlined here https://github.com/terraform-google-modules/terraform-example-foundation/pull/51
- Remove the main.tf which calls the sub components for the org stage, as we will work on a solution for using different state back-ends per component, rather than running all in one

cc @morgante - call out if you think any of this is off track

rjerrems:
Also on your comment around the 5GB of space in .terraform @mikelaramie - I suspect this is caused by gcloud module which is used in project factory among others. I recently ran into this with another repo with the same version of project factory we are using here.

Using `skip_gcloud_download` in the latest version of the project factory I was able to reduce the size of the `.terraform` directory from ~20GB to ~200mb -  so we should consider using the same approach here

PR #51, opened 2020-06-10
Split environment setup into separate subdirectories.

morgante:
This PR begins to move us closer to #47 by splitting org set up into separate subdirectories.

In particular, it introduces the `2-environments` directory which contains common env-specific configs. Modules and symlinks are used for reusing config.


rjerrems:
I like the general idea behind this and it will make adding new environments much easier. Also a good to make these more significant changes to develop until they are fully formed so master remains stable.

With the environments baseline, what other things will include? We could add VPC host project, shared secrets project etc?

morgante:
> With the environments baseline, what other things will include? We could add VPC host project, shared secrets project etc?

Exactly, I think most of the env-specific stuff can be moved here.

We'd also mirror this directory structure in the `networking` directory (but keep it separate).

rjerrems:
Ok cool sounds good to me - shall we move the other env specific resources across in the PR as well? We should also remove the monitoring projects from org as they are still there

morgante:
> Ok cool sounds good to me - shall we move the other env specific resources across in the PR as well?

We can leave that as something to do later.

> We should also remove the monitoring projects from org as they are still there

Done.

PR #53, opened 2020-06-13
refactor: move VPC host project into env module, add env folder

rjerrems:
Continuation of refactoring to make it easier to produce additional environment.


rjerrems:
FYI @mikelaramie

PR #55, opened 2020-06-18
Move secrets.tf to env_baseline module

bharathkkb:
- Move secrets.tf which spins up secrets project per env to env_baseline module


jonathan-moulds-sb:
@bharathkkb apologies if this is the wrong location and firstly thank you for the awesome work you are doing putting this amazing resource together.
But in your secrets.tf file your code still references the following `[name, application_name] = "env-secrets"`, should it be `[name, application_name] = "env-secrets-${var.env}"` to make it consistent with the networking and monitoring modules and environmental specific?

bharathkkb:
@jonathan-moulds-sb right, @langleyce will be clearing those up :)

Issue #52, opened 2020-06-11
Shared VPC Build


langleyce:
Why? To align existing foundations example with the best practice whitepaper guidance.
Updates:
- Per environment(x?)
  - Shared VPC
    - base
      - subnet region 1
      - subnet region 2
      - cloud router 1(x2)
      - cloud router 2(x2)
      - firewall rules (?)
      - cloud DNS
    - restricted
      - subnet region 1
      - subnet region 2
      - cloud router 1(x2)
      - cloud router 2(x2)
      - firewall rules (?)
      - cloud DNS    
- DNS Hub project(peered cloud DNS)
- Interconnect Attachment(x?)

morgante:
This should be done per environment. In particular:

- Create the host project(s) for each environment in 2-environments (see #51)
- Create the actual network resources for each environment in 3-networks

rjerrems:
I have also started part the work for the host project side of things here:

https://github.com/terraform-google-modules/terraform-example-foundation/pull/53

bharathkkb:
@daniel-cit I  think with the last `DNS Hub project` PR this is complete?

rjerrems:
Completed with this PR https://github.com/terraform-google-modules/terraform-example-foundation/pull/79

PR #57, opened 2020-06-25
Feature/updates standard shared vpc

amandakarina:
Hi @morgante and @rjerrems, could you, please, take a look at this proposal?

This pull request addresses the new resource naming format for 3-network module. 

This is the first part of a series of changes for issue #52. 

Changes in this PR:
- We changed all the names to follow the new pattern defined in the Security Best Practices whitepaper.
- We are pushing the naming definition of network and sub-network name inside the standard module.

We put the secondary ranges inside of the sub-network to link the two together, since there is no longer a sub-network name to link the sub-network with its secondary ranges.

Next pull request will add the second region and cloud router.

Any changes/suggestions?


bharathkkb:
@amandakarina this is a regression from  https://github.com/terraform-google-modules/terraform-example-foundation/pull/55
Can you rebase and fix those names https://github.com/terraform-google-modules/terraform-example-foundation/pull/55#issuecomment-646337435

amandakarina:
> @amandakarina this is a regression from #55
> Can you rebase and fix those names [#55 (comment)](https://github.com/terraform-google-modules/terraform-example-foundation/pull/55#issuecomment-646337435)

I will rebase after PR #58 be merged

3-networks/modules/standard_shared_vpc/variables.tf
```diff
@@ -14,41 +14,62 @@
...
 
-variable "network_name" {
+variable "environment_code" {
   type        = string
-  description = "Name for VPC."
+  description = "A short form of the folder level resources (environment) within the Google Cloud organization."
 }
 
-variable "default_region" {
+variable "nat_region" {
```
rjerrems, line 43: We should have Cloud NAT in each region, not a single region.

3-networks/modules/standard_shared_vpc/main.tf
```diff
@@ -111,3 +154,67 @@ resource "google_compute_router_nat" "default_nat" {
     enable = true
   }
 }
+
+module "region1_router1" {
+  source  = "terraform-google-modules/cloud-router/google"
+  name    = "cr-${local.vpc_name}-${local.subnets[0].subnet_region}-cr1"
+  project = var.project_id
+  network = module.main.network_name
+  region  = local.subnets[0].subnet_region
```
rjerrems, line 163: I think the regions should be provided explicitly as locals/variables, rather than being retrieved from subnets. (you might have multiple subnets in one region)

3-networks/modules/standard_shared_vpc/main.tf
```diff
@@ -18,36 +18,79 @@
...
+      subnet_flow_logs      = subnet.subnet_flow_logs
+      description           = subnet.description
+    }
+  ]
+
+  temp_secondary_ranges = [
+    for subnet in var.subnets : {
+      sub_name = "sb-${local.vpc_name}-${subnet.subnet_region}"
+      ranges = [
+        for range in subnet.secondary_ranges : {
```
rjerrems, line 42: same as comment above.

3-networks/main.tf
```diff
@@ -38,61 +51,81 @@ data "google_projects" "prod_host_project" {
...
+          ip_cidr_range = "192.168.32.0/23"
+        }
+      ]
     },
+    {
+      subnet_ip             = local.nonprod_cidrs_sub_network2
+      subnet_region         = var.subnet_region2
+      subnet_private_access = "true"
+      subnet_flow_logs      = "false"
+      bgp_asn               = [64516, 64517]
```
rjerrems, line 83: I don't think that BGP ASN should be an attribute of a subnet.

3-networks/main.tf
```diff
@@ -38,61 +51,81 @@ data "google_projects" "prod_host_project" {
...
+          ip_cidr_range = "192.168.96.0/19"
+        },
+        {
+          range_label   = "gke-svc"
+          ip_cidr_range = "192.168.128.0/23"
+        }
+      ]
     },
+    {
+      subnet_ip             = local.prod_cidrs_sub_network2
```
rjerrems, line 120: As mentioned above, I think the CIDR block should be here - otherwise it hurts readability

3-networks/modules/standard_shared_vpc/main.tf
```diff
@@ -18,36 +18,79 @@
   Shared VPC configuration
  *****************************************/
 
+locals {
+  vpc_type                = "shared"
+  vpc_label               = (var.vpc_label != "" ? var.vpc_label : "private")
```
rjerrems, line 23: is this required, should we just not use the naming in the whitepaper? We could also remove the variable then

3-networks/modules/standard_shared_vpc/variables.tf
```diff
@@ -14,41 +14,62 @@
...
-  description = "CIDR range for private service networking. Used for Cloud SQL and other managed services."
+variable "bgp_asn_nat" {
+  type        = number
+  description = "BGP ASN for NAT cloud routes."
 }
 
-variable "bgp_asn" {
-  type        = string
-  description = "BGP ASN for default cloud router."
+variable "subnets" {
```
rjerrems, line 58: I think we should use the same format as the network module, so that people can define subnets as documented in that module

https://github.com/terraform-google-modules/terraform-google-network/blob/master/variables.tf#L37

Otherwise this will be an extra layer than needs to be maintained

amandakarina:
> Thanks for the PR - I have added some comments, but at a high level the main things I think we need to look at:
> 
> * How can we make it easy to define the regions the customer wants to use, rather than pulling these out of the subnets that get created
> * For services that are required for both regions e.g Cloud NAT we need to create those for both regions
> * We need to revise the usage of BGP ASNs, these should not relate to the subnets but instead the cloud routers that get created
> * It would be good to remove some of the locals and explicitly define those values in the code that calls the module

Hi @rjerrems! We talked with Sean Leighton - sleighton about some of these questions:
* The Cloud  NAT  is actually a dual NIIC and it's only configured in one region, and the user can choose which region he wants to use.
* The Cloud Routers are related to the sub-networks, they should be created two per region per sub-network.
* The BGP ASN is a single one for all sub-networks that will connect using interconnect and another one for the Cloud router used in the NIIC.
* Also, the default firewall rules are workload dependents should not be enabled by default, so we are going to change the variable default value to false.

3-networks/modules/standard_shared_vpc/main.tf
```diff
@@ -111,3 +154,67 @@ resource "google_compute_router_nat" "default_nat" {
     enable = true
   }
 }
+
+module "region1_router1" {
+  source  = "terraform-google-modules/cloud-router/google"
+  name    = "cr-${local.vpc_name}-${local.subnets[0].subnet_region}-cr1"
```
bharathkkb, line 160: nit: version pin module here and throughout

3-networks/modules/standard_shared_vpc/variables.tf
```diff
@@ -14,41 +14,62 @@
  * limitations under the License.
  */
 
+/**
```
bharathkkb, line 17: nit: looks like repeated header license?

3-networks/variables.tf
```diff
@@ -26,7 +26,15 @@ variable "terraform_service_account" {
   description = "Service account email of the account to impersonate to run Terraform."
 }
 
-variable "default_region" {
-  description = "Default subnet region standard_shared_vpc currently only configures one region"
+variable "nat_region" {
+  description = "NAT region standard_shared_vpc"
```
bharathkkb, line 30: Is it possible to provide some sane defaults for these regions? I think overall we should standardize on a default region that can be followed through out.

rjerrems:
> > Thanks for the PR - I have added some comments, but at a high level the main things I think we need to look at:
> > 
> > * How can we make it easy to define the regions the customer wants to use, rather than pulling these out of the subnets that get created
> > * For services that are required for both regions e.g Cloud NAT we need to create those for both regions
> > * We need to revise the usage of BGP ASNs, these should not relate to the subnets but instead the cloud routers that get created
> > * It would be good to remove some of the locals and explicitly define those values in the code that calls the module
> 
> Hi @rjerrems! We talked with Sean Leighton - sleighton about some of these questions:
> 
> * The Cloud  NAT  is actually a dual NIIC and it's only configured in one region, and the user can choose which region he wants to use.
> * The Cloud Routers are related to the sub-networks, they should be created two per region per sub-network.
> * The BGP ASN is a single one for all sub-networks that will connect using interconnect and another one for the Cloud router used in the NIIC.
> * Also, the default firewall rules are workload dependents should not be enabled by default, so we are going to change the variable default value to false.

I don't quite understand why Cloud NAT is only configured in one region? This seems to defeat the purpose of having two regions in the first place. Is the whitepaper planning on pushing all egress traffic through a single region?

In my local testing, it does not seem that Cloud NAT even works between two regions so you would have to implement a proxy for that to work at all?

PR #60, opened 2020-06-27
Add restricted shared vpc

daniel-cit:
This PR has: 

- Creation of the **restricted shared vpc** module in `3-network`
- Creation of environments `dev`, `nonprod` e `prod` in `3-network`

TODOs:

- add Access Context Manager Admin - roles/accesscontextmanager.policyAdmin role the the Terraform service account
- enable `cloudresourcemanager.googleapis.com` and `accesscontextmanager.googleapis.com` apis in the "Restricted Shared VPC Host Project" of each env  after the code for the creation of these new projects is created.
- Merge the new version of the "Base Shared VPC Host Project" moving the module call to the envs.
- Update lookup labels for "Restricted Shared VPC Host Project" for each env after the code for the creation of these new projects is created.

One point I am not sure about is regarding the access level to user in the VPC service control configuration. 
Current one uses `members` but it cloud also be `ip_subnetworks`.


3-networks/modules/restricted_shared_vpc/main.tf
```diff
@@ -0,0 +1,210 @@
...
+      subnet_name           = "sb-${local.vpc_name}-${subnet.subnet_region}"
+      subnet_ip             = subnet.subnet_ip
+      subnet_region         = subnet.subnet_region
+      subnet_private_access = subnet.subnet_private_access
+      subnet_flow_logs      = subnet.subnet_flow_logs
+      description           = subnet.description
+    }
+  ]
+
+  temp_secondary_ranges = [
```
bharathkkb, line 38: Is there a reason why we need to do this manipulation as opposed to following the same interface as networks module?

3-networks/modules/restricted_shared_vpc/main.tf
```diff
@@ -0,0 +1,210 @@
...
+  project       = var.project_id
+  purpose       = "VPC_PEERING"
+  address_type  = "INTERNAL"
+  address       = element(split("/", var.private_service_cidr), 0)
+  prefix_length = element(split("/", var.private_service_cidr), 1)
+  network       = module.main.network_self_link
+}
+
+resource "google_service_networking_connection" "private_vpc_connection" {
+  provider                = google-beta
```
bharathkkb, line 100: Are we using any beta features in these resources where we are using the beta provider?

3-networks/modules/restricted_shared_vpc/main.tf
```diff
@@ -0,0 +1,210 @@
...
+  }
+}
+
+/************************************
+  Router to advertise shared VPC
+  subnetworks and Google Restricted API
+************************************/
+
+module "region1_router1" {
+  source  = "terraform-google-modules/cloud-router/google"
```
bharathkkb, line 149: version pin module here and throughout

3-networks/modules/restricted_shared_vpc/service_control.tf
```diff
@@ -0,0 +1,50 @@
...
+ * limitations under the License.
+ */
+
+locals {
+  parent_id = var.org_id
+  prefix    = "${var.environment_code}_${local.vpc_type}_${local.vpc_label}"
+}
+
+module "access_context_manager_policy" {
+  source      = "terraform-google-modules/vpc-service-controls/google"
```
bharathkkb, line 23: version pin here and throughout

3-networks/modules/restricted_shared_vpc/service_control.tf
```diff
@@ -0,0 +1,50 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+locals {
+  parent_id = var.org_id
```
bharathkkb, line 18: nit: not that familiar with VPC-SC but do we need local if this can only orgs ids?

3-networks/modules/restricted_shared_vpc/variables.tf
```diff
@@ -0,0 +1,110 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+variable "org_id" {
+  description = "Organization ID"
```
bharathkkb, line 18: nit: missing type

3-networks/modules/restricted_shared_vpc/firewall.tf
```diff
@@ -0,0 +1,73 @@
...
+
+// Allow SSH via IAP when using the allow-iap-ssh tag for Linux workloads.
+resource "google_compute_firewall" "allow_iap_ssh" {
+  count   = var.default_fw_rules_enabled ? 1 : 0
+  name    = "allow-iap-ssh"
+  network = module.main.network_name
+  project = var.project_id
+
+  // Cloud IAP's TCP forwarding netblock
+  source_ranges = concat(data.google_netblock_ip_ranges.iap_forwarders.cidr_blocks_ipv4)
```
bharathkkb, line 29: Do we need `concat` here and throughout if this is a single list?

3-networks/envs/nonprod/outputs.tf
```diff
@@ -0,0 +1,103 @@
...
+
+output "restricted_subnets_self_links" {
+  value       = module.restricted_shared_vpc.subnets_self_links
+  description = "The self-links of subnets being created"
+}
+
+output "restricted_subnets_secondary_ranges" {
+  value       = module.restricted_shared_vpc.subnets_secondary_ranges
+  description = "The secondary ranges associated with these subnets"
+}
```
bharathkkb, line 54: nit: fmt

3-networks/envs/prod/outputs.tf
```diff
@@ -0,0 +1,103 @@
...
+output "restricted_subnets_secondary_ranges" {
+  value       = module.restricted_shared_vpc.subnets_secondary_ranges
+  description = "The secondary ranges associated with these subnets"
+}
+output "restricted_allow_iap_ssh" {
+  value       = module.restricted_shared_vpc.allow_iap_ssh
+  description = "Firewall rule allow_iap_ssh created in the network"
+}
+
+output "restricted_allow_iap_rdp" {
```
bharathkkb, line 60: I am not sure if all these outputs are necessary, especially resources like FW. @langleyce is this a whitepaper specific thing?

3-networks/modules/restricted_shared_vpc/firewall.tf
```diff
@@ -0,0 +1,110 @@
...
+}
+
+/******************************************
+  Optional firewall rules
+ *****************************************/
+
+// Allow SSH via IAP when using the allow-iap-ssh tag for Linux workloads.
+resource "google_compute_firewall" "allow_iap_ssh" {
+  count   = var.optional_fw_rules_enabled ? 1 : 0
+  name    = "allow-iap-ssh"
```
rjerrems, line 61: Should we use the same naming convention for these?

3-networks/modules/restricted_shared_vpc/variables.tf
```diff
@@ -0,0 +1,120 @@
...
+variable "nat_enabled" {
+  type        = bool
+  description = "Toggle creation of NAT cloud router."
+  default     = false
+}
+
+variable "nat_bgp_asn1" {
+  type        = number
+  description = "BGP ASN for first NAT cloud routes."
+  default     = 0
```
rjerrems, line 46: Should we provide a default value for this? I would also prefer it was region1 so its clear which Cloud Router region / NAT it is associated with

3-networks/modules/restricted_shared_vpc/variables.tf
```diff
@@ -0,0 +1,120 @@
...
+}
+
+variable "nat_num_addresses1" {
```
rjerrems, line 49: similar to above, should be region1

3-networks/modules/restricted_shared_vpc/outputs.tf
```diff
@@ -0,0 +1,120 @@
...
+  value       = module.main.subnets_flow_logs
+  description = "Whether the subnets have VPC flow logs enabled"
+}
+
+output "subnets_secondary_ranges" {
+  value       = module.main.subnets_secondary_ranges
+  description = "The secondary ranges associated with these subnets"
+}
+
+output "allow_iap_ssh" {
```
rjerrems, line 62: As @bharathkkb mentioned, I don't think we need all these outputs - we should focus on adding the outputs that might be useful / required elsewhere. Also, because this in in a count I am not sure they will work

3-networks/modules/restricted_shared_vpc/nat.tf
```diff
@@ -0,0 +1,89 @@
...
+  name    = "ca-${local.vpc_name}-${var.subnets[0].subnet_region}-${count.index}"
+  region  = var.subnets[0].subnet_region
+}
+
+resource "google_compute_router_nat" "egress_nat1" {
+  count                              = var.nat_enabled ? 1 : 0
+  name                               = "rn-${local.vpc_name}-${var.subnets[0].subnet_region}-egress"
+  project                            = var.project_id
+  router                             = google_compute_router.nat_router1.0.name
+  region                             = var.subnets[0].subnet_region
```
rjerrems, line 45: can we get these regions from variables directly instead? i.e
var.region1
With the current implementation, it might have unintended consequences if customers add additional subnets

3-networks/envs/prod/outputs.tf
rjerrems: agree with this, would be good to limit these as excessive outputs make it difficult to refactor

3-networks/envs/dev/outputs.tf
```diff
@@ -0,0 +1,104 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/*********************
+ Restricted Outputs
```
rjerrems, line 18: Same comment about excessive outputs

3-networks/modules/restricted_shared_vpc/variables.tf
daniel-cit: NAT is now disable by the default and all NAT related variables have become optional. 
Adding a better description for it.

3-networks/modules/restricted_shared_vpc/main.tf
rjerrems: This is likely a relic of the past, from when this resource was in beta

3-networks/modules/restricted_shared_vpc/main.tf
```diff
@@ -0,0 +1,152 @@
...
+      {
+        name              = "rt-${local.vpc_name}-1000-egress-internet-default"
+        description       = "Tag based route through IGW to access internet"
+        destination_range = "0.0.0.0/0"
+        tags              = "egress-internet"
+        next_hop_internet = "true"
+        priority          = "1000"
+      },
+      {
+        name              = "rt-${local.vpc_name}-1000-all-default-windows-kms"
```
rjerrems, line 59: nit: this probably should not be bundled with NAT as windows activation is separate. If I recall correctly, it can still work with Google private access enabled, without NAT.

3-networks/modules/restricted_shared_vpc/nat.tf
```diff
@@ -0,0 +1,89 @@
...
+
+resource "google_compute_router" "nat_router_region1" {
+  count   = var.nat_enabled ? 1 : 0
+  name    = "cr-${local.vpc_name}-${var.default_region1}-nat-router"
+  project = var.project_id
+  region  = var.default_region1
+  network = module.main.network_self_link
+
+  bgp {
+    asn = var.nat_bgp_asn_region1
```
rjerrems, line 29: If possible, we should use a single ASN for both NAT Cloud Routers to limit the number of inputs required.

3-networks/modules/restricted_shared_vpc/variables.tf
rjerrems: Thanks! So after some extra thought, we could make this one ASN to make it consistent with the way subnet cloud routers are currently being created? Not a blocker for merging.

rjerrems: Also I think it would be best to have a default value which is valid

3-networks/modules/restricted_shared_vpc/main.tf
daniel-cit: `provider = google-beta` removed from resources.

0-bootstrap/main.tf
```diff
@@ -52,6 +52,35 @@ module "seed_bootstrap" {
...
+    "iam.googleapis.com",
+    "admin.googleapis.com",
+    "appengine.googleapis.com",
+    "storage-api.googleapis.com",
+    "monitoring.googleapis.com",
+    "accesscontextmanager.googleapis.com"
+  ]
+
+  sa_org_iam_permissions = [
+    "roles/accesscontextmanager.policyAdmin",
```
daniel-cit, line 73: @rjerrems this role and enabling the "accesscontextmanager.googleapis.com" were missing from the PR.
They are necessary for the VPC service Control configuration.

rjerrems: Ok sounds good to me

rjerrems:
Thanks for the updates @daniel-cit - looking good from my perspective. Call out if you see anything problematic @morgante / @bharathkkb , otherwise we can merge tomorrow.

PR #65, opened 2020-06-30
Adds integration test framework for e2e tests

bharathkkb:
- Adds test setup
- Adds fixtures for following stages
    - bootstrap
    - org
    - env

Todo:

- [x]  CB yaml

- [x] CB Trigger https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/pull/734

### DIFF
--- a/0-bootstrap/main.tf
+++ b/0-bootstrap/main.tf
@@ -43,8 +43,8 @@ resource "google_folder" "seed" {
 }
 
 module "seed_bootstrap" {
-  source = "github.com/terraform-google-modules/terraform-google-bootstrap"
-  # version                 = "~> 1.0"
+  source                  = "terraform-google-modules/bootstrap/google"
+  version                 = "~> 1.2"
   org_id                  = var.org_id
   folder_id               = google_folder.seed.id
   billing_account         = var.billing_account
@@ -90,8 +90,8 @@ module "seed_bootstrap" {
 }
 
 module "cloudbuild_bootstrap" {
-  source = "github.com/terraform-google-modules/terraform-google-bootstrap//modules/cloudbuild"
-  # version                 = "~> 1.0"
+  source                  = "terraform-google-modules/bootstrap/google//modules/cloudbuild"
+  version                 = "~> 1.2"
   org_id                  = var.org_id
   folder_id               = google_folder.seed.id
   billing_account         = var.billing_account
--- a/0-bootstrap/main.tf
+++ b/0-bootstrap/main.tf
@@ -43,15 +43,19 @@ resource "google_folder" "seed" {
 }
 
 module "seed_bootstrap" {
-  source                  = "terraform-google-modules/bootstrap/google"
-  version                 = "~> 1.0"
+  source = "github.com/terraform-google-modules/terraform-google-bootstrap"
+  # version                 = "~> 1.0"
   org_id                  = var.org_id
   folder_id               = google_folder.seed.id
   billing_account         = var.billing_account
   group_org_admins        = var.group_org_admins
   group_billing_admins    = var.group_billing_admins
   default_region          = var.default_region
+  org_project_creators    = var.org_project_creators
   sa_enable_impersonation = true
+  parent_folder           = var.parent_folder == "" ? "" : local.parent
+  skip_gcloud_download    = var.skip_gcloud_download
+
   activate_apis = [
     "serviceusage.googleapis.com",
     "servicenetworking.googleapis.com",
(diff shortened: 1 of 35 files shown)


rjerrems:
Overall structure is looking good from my point of view @bharathkkb

1-org/org_policy.tf
```diff
@@ -149,9 +149,13 @@ module "org_enforce_bucket_level_access" {
...
+  byte_length = 4
+}
+
 module "access_context_manager_policy" {
   source      = "terraform-google-modules/vpc-service-controls/google"
   version     = "~> 2.0.0"
-  parent_id   = local.organization_id
-  policy_name = "default policy"
+  parent_id   = var.org_id
+  policy_name = "default policy-${random_id.suffix.hex}"
```
bharathkkb, line 160: @rjerrems I am having some test failures [1] for this `access_context_manager_policy` resource. Looks like in one of the previous runs [2] a destroy step failed leaving this behind. I initially thought  I could bypass with a rand suffix, but it doesnt seem dependent on the policy. Any ideas on how we can prevent this in the future or clean up? 

[1]: https://pantheon.corp.google.com/cloud-build/builds/9cfa6a60-7055-4826-b5f7-57fd5fd26b0e;step=5?project=cloud-foundation-cicd
[2]: https://pantheon.corp.google.com/cloud-build/builds/fd9ecc79-f5aa-4c72-bc82-1f6ce6fb095f?project=cloud-foundation-cicd

rjerrems: Yeah I think we need to make sure this resource is cleaned up, even if it requires some gcloud commands. An org can only have one access context manager policy from my testing with gcloud commans

0-bootstrap/main.tf
```diff
@@ -43,15 +43,18 @@ resource "google_folder" "seed" {
 }
 
 module "seed_bootstrap" {
-  source                  = "terraform-google-modules/bootstrap/google"
-  version                 = "~> 1.0"
+  source = "github.com/terraform-google-modules/terraform-google-bootstrap"
+  # version                 = "~> 1.0"
```
rjerrems, line 47: is this supposed to be commented out?

bharathkkb: @rjerrems this was temporary as I was waiting for the new 1.2 bootstrap release, will revert.

1-org/org_policy.tf
bharathkkb: gotcha, sounds good.

bharathkkb:
@rjerrems this should be ready for review, I will add some kitchen tests in a subsequent PR.

0-bootstrap/main.tf
rjerrems: It looks like this is still commented out?

1-org/org_policy.tf
rjerrems, line 160: nit: the random suffice is not really required for this resource given that only one can exist in an org

build/int.cloudbuild.yaml
```diff
@@ -1,4 +1,4 @@
-# Copyright 2020 Google LLC
```
rjerrems, line 1: nit: this should be 2020

test/fixtures/org/main.tf
```diff
@@ -0,0 +1,28 @@
...
+ * limitations under the License.
+ */
+
+module "test" {
+  source                    = "../../../1-org"
+  parent_folder             = var.parent_folder
+  org_id                    = var.org_id
+  billing_account           = var.billing_account
+  terraform_service_account = var.terraform_sa_email
+  default_region            = "us-east4"
```
rjerrems, line 23: nit: it would be good to have some consistency with the regions used

build/int.cloudbuild.yaml
rjerrems: There are a few instances of this FYI

0-bootstrap/main.tf
bharathkkb: just fixed :)

1-org/org_policy.tf
bharathkkb: removed

build/int.cloudbuild.yaml
bharathkkb: done

test/fixtures/org/main.tf
bharathkkb: using `us-central1` for consistency with other fixtures

bharathkkb:
@rjerrems I have addressed the comments but tests are failing due to a [Github packages outage](https://www.githubstatus.com/).  I will re run them when its back.

rjerrems:
I just triggered the builds again, looks like there may be some errors based on the region for BigQuery - I don't think you can specify us-central1

bharathkkb:
@rjerrems issue was here https://github.com/terraform-google-modules/terraform-example-foundation/pull/65#discussion_r453456499 thats probably why i went with `"us-east4"`. Reverting to that now.

PR #66, opened 2020-07-03
Adds SCC Alerting project and SCC Notification

amandakarina:
Hi @rjerrems, @morgante and @bharathkkb 

This pull requests adds a project, topic, subscription and a SCC Notification.

Could you please take a look? :smile:

Notes:
 - We have some questions regarding the SCC filter variable, since double quotes must be escaped. Can you advise us how to proceed? Example bellow.
- ~About the SCC Notification host project, we put it under logs folders. But since this folders are only logical, should the parent be the organization/folder provided on variables.tf?~

**Filter example:**

 The Security Command Center API expects escaped double quotes:
```
"state=\"ACTIVE\""
```

In the value of the variable scc_notification_filter we need to escape the original escape and the double quotes.
```
"state=\\\"ACTIVE\\\""
```

- @bharathkkb thanks for the idea for the skip_download in your pull request :smiley:


1-org/scc_notification.tf
```diff
@@ -0,0 +1,71 @@
...
+}
+
+
+module "scc_notification" {
+  source  = "terraform-google-modules/gcloud/google"
+  version = "~> 1.1.0"
+
+  platform = "linux"
+
+  additional_components = var.skip_gcloud_download ? [] : ["alpha"]
```
bharathkkb, line 53: I am not sure if we need this check? We would still need to make sure that `alpha` components are installed regardless?

1-org/scc_notification.tf
```diff
@@ -0,0 +1,71 @@
...
+    ttl = "300000.5s"
+  }
+}
+
+
+module "scc_notification" {
+  source  = "terraform-google-modules/gcloud/google"
+  version = "~> 1.1.0"
+
+  platform = "linux"
```
bharathkkb, line 51: nit: unnecessary as gcloud module defaults to linux

1-org/scc_notification.tf
```diff
@@ -0,0 +1,71 @@
...
+  name    = "top-scc-notification"
+  project = module.scc_notifications.project_id
+}
+
+resource "google_pubsub_subscription" "scc_notification_subscription" {
+  name    = "sub-scc-notification"
+  topic   = google_pubsub_topic.scc_notification_topic.name
+  project = module.scc_notifications.project_id
+
+  message_retention_duration = "1200s"
```
bharathkkb, line 36: Are we giving opinionated best values for constants like `message_retention_duration` `retain_acked_messages` `ack_deadline_seconds ` `expiration_policy` etc from an SCC pov? Otherwise I think we should just leave these to provider defaults.

0-bootstrap/main.tf
```diff
@@ -52,6 +52,34 @@ module "seed_bootstrap" {
...
+  sa_org_iam_permissions = [
+    "roles/billing.user",
+    "roles/compute.networkAdmin",
+    "roles/compute.xpnAdmin",
+    "roles/iam.securityAdmin",
+    "roles/iam.serviceAccountAdmin",
+    "roles/logging.configWriter",
+    "roles/orgpolicy.policyAdmin",
+    "roles/resourcemanager.folderAdmin",
+    "roles/pubsub.admin",
```
rjerrems, line 80: Pubsub permissions should be granted on the SCC project, not on the org (this list of permissions is granted across the org)

1-org/scc_notification.tf
```diff
@@ -0,0 +1,71 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/******************************************
+  SCC Notification
+*****************************************/
+
+locals {
+  description = "SCC Notification for all active findings"
```
rjerrems, line 22: Given this is only used once, you can probably just put it inline rather than storing as a local

1-org/scc_notification.tf
rjerrems: Agree with this, unless these are the values from documentation / white paper then we should use provider defaults

1-org/variables.tf
```diff
@@ -67,6 +67,23 @@ variable "data_access_table_expiration_ms" {
...
+variable "skip_gcloud_download" {
+  description = "Whether to skip downloading gcloud (assumes gcloud is already available outside the module)"
+  type        = bool
+  default     = true
+}
+
+variable "scc_notification_filter" {
+  description = "Filter used to SCC Notification, you can see more details how to create filters in https://cloud.google.com/security-command-center/docs/how-to-api-filter-notifications#create-filter"
+  type        = string
+  default     = "state=\\\"ACTIVE\\\""
```
rjerrems, line 84: Are all of these backslashes required? The documentation seems to only require one

1-org/scc_notification.tf
amandakarina: Hi @bharathkkb

If we skip download and force the `additional_components = ['alpha']` we got this error:

```
Error: Error running command 'gcloud components install alpha --quiet': exit status 1. Output: ERROR: (gcloud.components.install) 
You cannot perform this action because the Cloud SDK component manager 
is disabled for this installation. You can run the following command 
to achieve the same result for this installation: 

sudo apt-get install google-cloud-sdk
```

So that's why the ternary was added. Any advice?

bharathkkb: @amandakarina are you running in a system where gcloud was installed using apt/yum? [Docs](https://cloud.google.com/sdk/docs/components#managing_cloud_sdk_components) suggest that may result in component manager being disabled. Could you test it inside the cft container?

amandakarina: Hi @bharathkkb running inside the cft container I've got no error.

1-org/projects.tf
```diff
@@ -75,3 +75,26 @@ module "org_secrets" {
...
+module "scc_notifications" {
+  source                      = "terraform-google-modules/project-factory/google"
+  version                     = "~> 8.0"
+  random_project_id           = "true"
+  impersonate_service_account = var.terraform_service_account
+  default_service_account     = "depriviledge"
+  name                        = "prj-scc-notifications"
+  org_id                      = var.org_id
+  billing_account             = var.billing_account
+  folder_id                   = var.parent_folder
```
daniel-cit, line 92: According to the whitepaper the parent folder should be  `google_folder.common.id`

1-org/scc_notification.tf
bharathkkb: @amandakarina wouldnt it just say? 
```
[root@5602acbb5270 workspace]# gcloud components install alpha
All components are up to date.
```
If you are still running into it, lets keep the ternary. We can revisit later if needed.
I think the check you pointed our is great, will also look into fixing this in the gcloud module.

morgante:
@rjerrems Can you give this another pass/final review?

PR #71, opened 2020-07-07
Adds interconnect project

amandakarina:
Hello folks!

This pull requests creates the interconnect host project.

Could you please, take a look?

Thanks!

### DIFF
--- a/1-org/projects.tf
+++ b/1-org/projects.tf
@@ -75,3 +75,25 @@ module "org_secrets" {
     application_name = "org-secrets"
   }
 }
+
+/******************************************
+  Project for Interconnect
+*****************************************/
+
+module "interconnect" {
+  source                      = "terraform-google-modules/project-factory/google"
+  version                     = "~> 8.0"
+  random_project_id           = "true"
+  impersonate_service_account = var.terraform_service_account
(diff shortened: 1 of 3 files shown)


1-org/projects.tf
```diff
@@ -75,3 +75,25 @@ module "org_secrets" {
...
+module "interconnect" {
+  source                      = "terraform-google-modules/project-factory/google"
+  version                     = "~> 8.0"
+  random_project_id           = "true"
+  impersonate_service_account = var.terraform_service_account
+  default_service_account     = "depriviledge"
+  name                        = "prj-interconnect"
+  org_id                      = var.org_id
+  billing_account             = var.billing_account
+  folder_id                   = var.parent_folder
```
daniel-cit, line 92: According to the whitepaper the parent folder should be  `google_folder.common.id`

1-org/variables.tf
```diff
@@ -67,6 +67,12 @@ variable "data_access_table_expiration_ms" {
   default     = 2592000000
 }
 
+variable "skip_gcloud_download" {
```
bharathkkb, line 70: @amandakarina is there a reason why we are calling out the need for `gcloud alpha module`?

PR #72, opened 2020-07-07
Add private(base) shared VPC to each environment.

daniel-cit:
Changes in this pull request

- deleting old code from 3-networks.
- deleting creation of old shared VPC projects from `2-environmet`.
- Calling new base shared VPC module in each environment.
- moving creation of Access Context Manager default policy to `1-org` since the policy is uniq for the organization.
- adding terraform.example.tfvars and versions.tf to each environment
- updating README.md for each environment

@bharathkkb @rjerrems @morgante cçoud you please take a look at these changes


daniel-cit:
@rjerrems this PR should be the last one for the new basic part of the shared VPCs.

Next PRs will be related to interconnects and VPNs

PR #74, opened 2020-07-08
Create dedicated interconnect module.

daniel-cit:
Creation of  dedicated interconnect module in `3-networks`.

TODOs:

- [x] add  commented call in all the environments for both private and restricted shared VPC after https://github.com/terraform-google-modules/terraform-example-foundation/pull/72 is merged.
- [x] update README.md file in the new module and in each environment to presente the new module as the recommended option for onprem connection.

Depends on:

- https://github.com/terraform-google-modules/terraform-example-foundation/pull/72
- https://github.com/terraform-google-modules/terraform-example-foundation/pull/71

### DIFF
--- a/3-networks/modules/dedicated_interconnect/README.md
+++ b/3-networks/modules/dedicated_interconnect/README.md
@@ -6,7 +6,7 @@ This module implementes the recomendation proposed in [Establishing 99.99% Avail
 
 1. Execution of at least one of the environments in `3-networks/env/`.
 1. Creation on two Cloud Routers for each Region. Both the `Restricted shared VPC` and the `Standard shared VPC` modules create two Cloud Routers for each Region that should be used as input for this module. **This module only works with two regions.**
-1. Provisioning of four [dedicated interconnects](https://cloud.google.com/network-connectivity/docs/interconnect/concepts/dedicated-overview), Two for each region.
+1. Provisioning of four [dedicated interconnects](https://cloud.google.com/network-connectivity/docs/interconnect/concepts/dedicated-overview) in the `prj-interconnect` project created in step `1-org` under folder `fldr-common`.
 
 ## Usage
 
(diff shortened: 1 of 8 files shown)


3-networks/modules/dedicated_interconnect/main.tf
```diff
@@ -0,0 +1,111 @@
...
+
+locals {
+  interconnect_project_id = data.google_projects.interconnect_project.projects[0].project_id
+}
+
+data "google_projects" "interconnect_project" {
+  filter = "labels.application_name=prj-interconnect"
+}
+
+module "interconnect_attachment1_region1" {
```
bharathkkb, line 25: Not too familiar with this module but does it make sense to have any outputs for these?

3-networks/modules/dedicated_interconnect/main.tf
```diff
@@ -0,0 +1,111 @@
...
+    peer_ip_address = var.peer_ip_address
+    peer_asn        = var.peer_asn
+  }
+}
+
+module "interconnect_attachment2_region1" {
+  source  = "terraform-google-modules/cloud-router/google//modules/interconnect_attachment"
+  version = "~> 0.2.0"
+
+  name    = "vl-${var.interconnect_location2_region1}-${var.vpc_name}-${var.default_region1}-cr1"
```
rjerrems, line 51: should this be cr2?

3-networks/modules/dedicated_interconnect/main.tf
```diff
@@ -0,0 +1,111 @@
...
+    peer_ip_address = var.peer_ip_address
+    peer_asn        = var.peer_asn
+  }
+}
+
+module "interconnect_attachment2_region2" {
+  source  = "terraform-google-modules/cloud-router/google//modules/interconnect_attachment"
+  version = "~> 0.2.0"
+
+  name    = "vl-${var.interconnect_location2_region2}-${var.vpc_name}-${var.default_region2}-cr1"
```
rjerrems, line 95: should this also be cr2?

rjerrems:
Overall looking good aside from a couple of resource names. Should we also have some docs about what needs to happen before the attachments are created or is that in the whitepaper?

3-networks/modules/dedicated_interconnect/README.md
```diff
@@ -0,0 +1,25 @@
+<!-- BEGINNING OF PRE-COMMIT-TERRAFORM DOCS HOOK -->
```
rjerrems, line 1: We should provide a brief overview of what the module is designed to do, even if only a few sentences.

3-networks/modules/dedicated_interconnect/main.tf
daniel-cit: adding to the module output the same outputs of the underlying module

3-networks/modules/dedicated_interconnect/main.tf
daniel-cit: fixed

3-networks/modules/dedicated_interconnect/main.tf
daniel-cit: fixed

3-networks/modules/dedicated_interconnect/README.md
daniel-cit: Added an explanation and link to public documentation and also added commented example calls from the environments.

3-networks/modules/dedicated_interconnect/outputs.tf
```diff
@@ -0,0 +1,55 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+output "interconnect_attachment1_region1" {
+  value       = module.interconnect_attachment1_region1.attachment
+  description = "The created attachment"
```
rjerrems, line 19: nit: these should be more specific. I.e "The interconnect attachment 1 for region 1"

PR #75, opened 2020-07-08
[DRAFT] High availability VPN module

amandakarina:
Creation of High availability VPN module in 3-networks.

TODOs:

- [x] add example of how to call vpn module
- [x] update README.md file in the new module


3-networks/modules/vpn-ha/variables.tf
```diff
@@ -0,0 +1,96 @@
...
+  type        = string
+  description = "Remote 1 IP range"
+}
+
+variable "remote1_secret" {
+  type        = string
+  description = "Remote 1 secret name"
+}
+
+variable "region1_router1_name" {
```
morgante, line 78: We shouldn't need to input this.

3-networks/modules/vpn-ha/variables.tf
```diff
@@ -0,0 +1,96 @@
...
+  type        = string
+  description = "Default region 2 for Cloud Routers"
+}
+
+variable "network_name" {
+  type        = string
+  description = "Network Name"
+}
+
+variable "vpc_name" {
```
morgante, line 32: Why is there vpc_name *and* network_name?

3-networks/modules/vpn-ha/variables.tf
```diff
@@ -0,0 +1,96 @@
...
+
+variable "on_prem_ip_address" {
+  type        = string
+  description = "On-Prem IP address"
+}
+
+variable "bgp_asn" {
+  type        = number
+  description = "BGP ASN for cloud routes."
+  default     = 0
```
morgante, line 45: This doesn't seem like a good default.

3-networks/modules/vpn-ha/README.md
```diff
@@ -0,0 +1,23 @@
+<!-- BEGINNING OF PRE-COMMIT-TERRAFORM DOCS HOOK -->
```
rjerrems, line 1: We should add an overview of what the module does here, even if it is only a couple of sentences.

3-networks/modules/vpn-ha/main.tf
```diff
@@ -0,0 +1,199 @@
...
+  interconnect_project_id = data.google_projects.interconnect_project.projects[0].project_id
+}
+
+data "google_projects" "interconnect_project" {
+  filter = "labels.application_name=prj-interconnect"
+}
+
+module "vpn_ha_region1_router1" {
+  source     = "terraform-google-modules/vpn/google//modules/vpn_ha"
+  project_id = local.interconnect_project_id
```
rjerrems, line 31: I believe the VPN has to be in the same project as the VPC so this may need to be a parameter.

3-networks/modules/vpn-ha/main.tf
```diff
@@ -0,0 +1,199 @@
...
+  region     = var.default_region1
+  network    = var.network_name
+  name       = "vpn-${var.vpc_name}-${var.default_region1}-cr1"
+  peer_external_gateway = {
+    redundancy_type = "SINGLE_IP_INTERNALLY_REDUNDANT"
+    interfaces = [{
+      id         = 0
+      ip_address = var.on_prem_ip_address
+
```
rjerrems, line 40: Can probably remove the whitespace here

3-networks/modules/vpn-ha/variables.tf
```diff
@@ -0,0 +1,96 @@
...
+}
+
+variable "remote1_range" {
+  type        = string
+  description = "Remote 1 IP range"
+}
+
+variable "remote1_secret" {
+  type        = string
+  description = "Remote 1 secret name"
```
rjerrems, line 75: Is this the name of the secret or the actual secret? We should consider some options / instructions to help make sure these are not stored in plain text in the tfvars / code.

3-networks/modules/vpn-ha/variables.tf
rjerrems: +1 should we even have a default given the value is required?

3-networks/modules/vpn-ha/main.tf
```diff
@@ -0,0 +1,199 @@
...
+  source     = "terraform-google-modules/vpn/google//modules/vpn_ha"
+  project_id = local.interconnect_project_id
+  region     = var.default_region2
+  network    = var.network_name
+  name       = "vpn-${var.vpc_name}-${var.default_region2}-cr1"
+  peer_external_gateway = {
+    redundancy_type = "SINGLE_IP_INTERNALLY_REDUNDANT"
+    interfaces = [{
+      id         = 0
+      ip_address = var.on_prem_ip_address
```
rjerrems, line 125: Across two regions, we would probably expect to have two different on-prem IPs rather than one right?

rjerrems:
Heading in the right direction @amandakarina thanks! I have added some comments

3-networks/modules/vpn-ha/main.tf
```diff
@@ -0,0 +1,209 @@
...
+    interfaces = [{
+      id         = 0
+      ip_address = var.on_prem_router_ip_address1
+      },
+      {
+        id         = 1
+        ip_address = var.on_prem_router_ip_address2
+    }]
+  }
+  router_name = "cr-${local.vpc_name}-${var.default_region1}-cr1"
```
daniel-cit, line 44: The `router_name` could be received as an explicit variable since it should exist and have been already created in the same context this module will be called.

3-networks/modules/vpn-ha/main.tf
```diff
@@ -0,0 +1,209 @@
...
+ * limitations under the License.
+ */
+
+/******************************************
+  HA VPN configuration
+ *****************************************/
+
+locals {
+  vpc_name     = "${var.environment_code}-shared-${var.vpc_label}"
+  network_name = "vpc-${local.vpc_name}"
```
daniel-cit, line 23: Both `vpc_name`  and `network_name` could be variables since they are defined in the context the module is called.

3-networks/modules/vpn-ha/README.md
```diff
@@ -0,0 +1,100 @@
...
+  # tunnel 2
+  bgp_peer_address2 = "169.254.4.1"
+  bgp_peer_range2 = "169.254.4.2/30"
+
+  # tunnel 3
+  bgp_peer_address3 = "169.254.6.1"
+  bgp_peer_range3 = "169.254.6.2/30"
+
+  # tunnel 4
+  bgp_peer_address4 = "169.254.8.1"
```
rjerrems, line 52: I think it would be better for this to reference the region, otherwise its unclear reading the code which region / tunnel it belongs too

3-networks/modules/vpn-ha/README.md
```diff
@@ -0,0 +1,100 @@
...
+## Prerequisites
+
+1. Execution of at least one of the environments in `3-networks/env/`.
+1. Creation on two Cloud Routers for each Region. Both the `Restricted shared VPC`
+and the `Standard shared VPC` modules create two Cloud Routers for each Region
+that should be used as input for this module. **This module only works with two regions.**
+
+## Usage
+
+For each environmnet and for each network, you should add the following peace of code in the
```
rjerrems, line 15: nit: replace peace with piece or snippet

rjerrems: nit: environmnet should also be environment

3-networks/modules/vpn-ha/variables.tf
amandakarina: Hi @morgante 
We are inputting the routers since we need to connect on-prem and the cloud routers have already been created in 3-network (standard and restricted vpc) for use in the interconnect.
If we don't inform a router name, other routers will be automatically created, is this your suggestion?

3-networks/modules/vpn-ha/variables.tf
amandakarina: Only `vpc_name` will gonna be input.

3-networks/modules/vpn-ha/variables.tf
amandakarina: Hi @rjerrems, we are going to use https://www.terraform.io/docs/providers/google/d/kms_secret.html to solve this:
- Adding the kms key ring and key in `2-environments/modules/env_baseline/secrets.tf` (a flag will be added since vpn is optional)
- Adding instruction on README to use gcloud to encrypt the secret
- Adding a data on `3-networks/modules/vpn-ha/main.tf` to retrieve the secret value
- Change the `3-networks/modules/vpn-ha/main.tf` to use the data where vpn secret is been used.
- Change the variable `bgp_secret` to `encrypted_bpg_secret` (the ciphertext, encrypted secret) in modules/vpn-ha

@morgante and @bharathkkb what do you think about it? Any suggestions?

3-networks/modules/vpn-ha/main.tf
amandakarina: Yes, already fixed :)

3-networks/modules/vpn-ha/main.tf
amandakarina: Yes :) Already fixed

3-networks/modules/vpn-ha/variables.tf
morgante: Ok.

3-networks/modules/vpn-ha/variables.tf
rjerrems: Is this going to work with our pipeline approach?

3-networks/modules/vpn-ha/README.md
amandakarina: Agree!

3-networks/modules/vpn-ha/README.md
amandakarina: Done!

amandakarina:
Waiting solution about how to handle PSK secret.

3-networks/envs/dev/terraform.tfvars
```diff
@@ -0,0 +1,25 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+org_id = "1055058813388"
+
+terraform_service_account = "org-terraform@cft-seed-62c3.iam.gserviceaccount.com"
```
rjerrems, line 19: I am assuming these should not be here and should be generic examples instead?

rjerrems:
I have added basic support for secret manager for VPN module.. let me know what you guys think, if we aren't happy with the approach we can revert to previous.

PR #77, opened 2020-07-10
Jenkins agent 2nd PR

caleonardo:
This 2nd PR creates the GCS bucket, KMS configuration and IAM permissions that Jenkins service account needs to impersonate the Terraform service account

TODO for next PRs:
- Jenkins GCE startup script to install java
- Jenkinsfile with pipeline
- VPN configuration

### DIFF
--- a/0-bootstrap/modules/jenkins/main.tf
+++ b/0-bootstrap/modules/jenkins/main.tf
@@ -1,0 +1,118 @@
+/**
+ * Copyright 2019 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+locals {
+  cicd_project_name           = format("%s-%s", var.project_prefix, "cicd")
+  impersonation_enabled_count = var.sa_enable_impersonation ? 1 : 0
+  activate_apis               = distinct(var.activate_apis)
+  jenkins_gce_fw_tags         = ["ssh-jenkins-agent"]
+}
+
+/******************************************
+  CICD project
+*******************************************/
+module "cicd_project" {
+  source                      = "terraform-google-modules/project-factory/google"
+  version                     = "~> 7.0" // TODO(caleonardo): use latest current versions so that we are up to date
+  name                        = local.cicd_project_name
+  random_project_id           = true
+  disable_services_on_destroy = false
+  folder_id                   = var.folder_id
+  org_id                      = var.org_id
+  billing_account             = var.billing_account
+  activate_apis               = local.activate_apis
+  labels                      = var.project_labels
+}
+
+/******************************************
+  Jenkins Agent GCE instance
+*******************************************/
+resource "google_service_account" "jenkins_agent_gce_sa" {
+  project      = module.cicd_project.project_id
+  account_id   = var.jenkins_sa_email
+  display_name = "CFT Jenkins Agent GCE custom Service Account"
+}
+
+resource "google_compute_instance" "jenkins_agent_gce_instance" {
+  project      = module.cicd_project.project_id
+  name         = var.jenkins_agent_gce_name
+  machine_type = var.jenkins_agent_gce_machine_type
+  zone         = "${var.default_region}-a"
+
+  tags = local.jenkins_gce_fw_tags
+
+  boot_disk {
+    initialize_params {
+      image = "debian-cloud/debian-9"
+    }
+  }
+
+  network_interface {
+    network = google_compute_network.jenkins_agents.self_link
+
+    access_config {
+      // Ephemeral IP
+      // TODO(caleonardo): This must not have an external IP at all - Only use for testing while developing
+    }
+  }
+
+  // Adding ssh public keys to the metadata, so the Jenkins Master can connect
+  metadata = {
+    enable-oslogin = "false"
+    ssh-keys       = "${var.jenkins_agent_gce_ssh_user}:${file(var.jenkins_agent_gce_ssh_pub_key_file)}"
+  }
+
+  // TODO(caleonardo): Setup the Java installation here for the Jenkins Agent
+  metadata_startup_script = "echo hi > /test.txt"
+
+  service_account {
+    email = google_service_account.jenkins_agent_gce_sa.email
+    scopes = [
+      // TODO(caleonardo): These scopes will need to change.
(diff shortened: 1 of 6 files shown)


0-bootstrap/modules/jenkins/main.tf
```diff
@@ -37,13 +42,20 @@ module "cicd_project" {
   labels                      = var.project_labels
 }
 
+resource "google_project_service" "jenkins_apis" {
```
bharathkkb, line 45: you can use `activate_apis ` with `cicd_project` PF.

0-bootstrap/modules/jenkins/main.tf
```diff
@@ -112,7 +125,147 @@ resource "google_compute_firewall" "allow_ssh_to_jenkins_agent_fw" {
...
+  project = module.cicd_project.project_id
+  role    = "roles/viewer"
+  member  = "group:${var.group_org_admins}"
+}
+
+/******************************************
+  Jenkins Artifact bucket
+*******************************************/
+
+resource "google_storage_bucket" "gcs_jenkins_artifacts" {
```
bharathkkb, line 160: what gcs artifacts are expected here?

0-bootstrap/modules/jenkins/main.tf
```diff
@@ -112,7 +125,147 @@ resource "google_compute_firewall" "allow_ssh_to_jenkins_agent_fw" {
...
+
+resource "google_kms_crypto_key_iam_binding" "jenkins_crypto_key_decrypter" {
+  crypto_key_id = google_kms_crypto_key.tf_key.self_link
+  role          = "roles/cloudkms.cryptoKeyDecrypter"
+
+  members = [
+    "serviceAccount:${google_service_account.jenkins_agent_gce_sa.email}",
+    "serviceAccount:${var.terraform_sa_email}"
+  ]
+  depends_on = [
```
bharathkkb, line 207: explicit dependency can be removed if using `activate_apis` here and throughout

0-bootstrap/modules/jenkins/variables.tf
```diff
@@ -111,8 +137,32 @@ variable "sa_enable_impersonation" {
...
+  default     = {}
+}
+
 variable "folder_id" {
   description = "The ID of a folder to host this project"
   type        = string
   default     = ""
 }
+
+variable "terraform_version" {
```
bharathkkb, line 158: will these be used by the startup scripts for the agent?

caleonardo: Yes. I added a template to use this now. Thanks

0-bootstrap/modules/jenkins/main.tf
caleonardo: People can put here Jenkins build logs if they want to. It is optional. By default, any output from the Jenkins build will go to the GCE hard drive (`/home/jenkins`). Being optional, should we remove this then?

0-bootstrap/modules/jenkins/main.tf
caleonardo: `google_project_service.jenkins_apis` is used in multiple places, so it makes sense to keep it as a separate resource.

0-bootstrap/modules/jenkins/main.tf
caleonardo: Yes, this is removed now. Thank you.

caleonardo:
@bharathkkb - I think the PR #78 should be enough and includes the changes in this one...

0-bootstrap/modules/jenkins/main.tf
bharathkkb: @caleonardo I am not sure I follow. All the places that use `google_project_service.jenkins_apis` are to enforce explicit dependecy. If you use `activate_apis` this is also not necessary as it will be enforced implicitly.

0-bootstrap/modules/jenkins/main.tf
bharathkkb: I think this is a good to have, let keep it.

0-bootstrap/modules/jenkins/main.tf
rjerrems: FYI this is separate in the bootstrap module like this, because we had a race condition at the time with APIs not being enabled before we created repos etc. Not sure if this is still the case in newer versions of the project factory

0-bootstrap/modules/jenkins/variables.tf
```diff
@@ -86,20 +112,20 @@ variable "project_prefix" {
...
     "compute.googleapis.com",
     "logging.googleapis.com",
-    "bigquery.googleapis.com",
+    "bigquery.googleapis.com", // TODO(caleonardo): confirm if CICD Project needs BQ
     "cloudresourcemanager.googleapis.com",
     "cloudbilling.googleapis.com",
     "iam.googleapis.com",
     "admin.googleapis.com",
-    "appengine.googleapis.com",
+    "appengine.googleapis.com", // TODO(caleonardo): confirm if CICD Project needs GAE
```
rjerrems, line 128: I believe it does, I can't remember off the top of my head why

0-bootstrap/modules/jenkins/variables.tf
```diff
@@ -86,20 +112,20 @@ variable "project_prefix" {
...
+  description = "List of APIs to enable in the CICD project."
   type        = list(string)
 
   default = [
     "serviceusage.googleapis.com",
     "servicenetworking.googleapis.com",
     "compute.googleapis.com",
     "logging.googleapis.com",
-    "bigquery.googleapis.com",
+    "bigquery.googleapis.com", // TODO(caleonardo): confirm if CICD Project needs BQ
```
rjerrems, line 123: This is required to create the BQ log sinks

0-bootstrap/modules/jenkins/variables.tf
caleonardo: Great. Thank you

0-bootstrap/modules/jenkins/variables.tf
caleonardo: Great. Thank you.

0-bootstrap/modules/jenkins/main.tf
caleonardo: Got it. Fixed in jenkins-agent-3 https://github.com/terraform-google-modules/terraform-example-foundation/pull/78

caleonardo:
@bharathkkb and @rjerrems - Can I get this PR approved now so I can send the next one?

caleonardo:
Ok, I am closing this PR, since all the changes required here and more are already included in PR#78: https://github.com/terraform-google-modules/terraform-example-foundation/pull/78

0-bootstrap/modules/jenkins/main.tf
bharathkkb: @rjerrems I dont think there is any race condition now.

caleonardo: This is already fixed in PR 78 which is the valid one. this PR77 is closed now.

Thank you!

PR #78, opened 2020-07-11
Jenkins agent 3

caleonardo:
Hi @bharathkkb and @rjerrems - could you please review this PR and let me know any changes you need me to implemente?

Thank you.

TODO: VPN Configuration for the CICD project


caleonardo:
@morgante - Could you please help me with this failing test? - it says:
```
Running shellcheck
Checking file headers
1 files have incorrect boilerplate headers:
0-bootstrap/modules/jenkins/scripts/jenkins_gce_startup_script.sh
```

But I don't see what I should change in the the script. When I run the `terraform apply`, the startup script works just fine. 

What should I change in the file?

Thank you!

daniel-cit:
> @morgante - Could you please help me with this failing test? - it says:
> 
> ```
> Running shellcheck
> Checking file headers
> 1 files have incorrect boilerplate headers:
> 0-bootstrap/modules/jenkins/scripts/jenkins_gce_startup_script.sh
> ```
> 
> But I don't see what I should change in the the script. When I run the `terraform apply`, the startup script works just fine.
> 
> What should I change in the file?
> 
> Thank you!

@caleonardo You need to add the "Copyright  notice" section after the Bash Shebang like on this example :  https://github.com/terraform-google-modules/terraform-google-gcloud/blob/master/examples/dependency_example/scripts/script.sh

0-bootstrap/modules/jenkins/scripts/jenkins_gce_startup_script.sh
```diff
@@ -0,0 +1,31 @@
...
+
+echo "**** Startup Step 1/9: Update apt-get repositories. ****"
+sudo apt-get update
+
+echo "**** Startup Step 2/9: Install Java. ****"
+sudo apt-get install -y default-jdk
+
+echo "**** Startup Step 3/9: Set variables. ****"
+export TERRAFORM_DIR="/opt/terraform/"
+export TERRAFORM_VERSION="0.12.24"
```
bharathkkb, line 11: Shouldn't we be templating this based on
https://github.com/terraform-google-modules/terraform-example-foundation/pull/77#discussion_r453252131

0-bootstrap/modules/jenkins/scripts/jenkins_gce_startup_script.sh
```diff
@@ -0,0 +1,31 @@
...
+echo "**** Startup Step 6/9: Install unzip. ****"
+sudo apt-get install -y unzip
+
+echo "**** Startup Step 7/9: Unzip the Terraform download. ****"
+unzip "terraform_{$TERRAFORM_VERSION}_linux_amd64.zip"
+chmod 755 $TERRAFORM_DIR/terraform
+
+echo "**** Startup Step 8/9: Download and install the Terraform validator ****"
+gsutil cp gs://terraform-validator/releases/2019-04-04/terraform-validator-linux-amd64 .
+chmod 755 $TERRAFORM_DIR/terraform-validator-linux-amd64
```
bharathkkb, line 28: nit: might make sense to just rename to `terraform-validator`

0-bootstrap/modules/jenkins/scripts/jenkins_gce_startup_script.sh
```diff
@@ -0,0 +1,31 @@
...
+sudo apt-get update
+
+echo "**** Startup Step 2/9: Install Java. ****"
+sudo apt-get install -y default-jdk
+
+echo "**** Startup Step 3/9: Set variables. ****"
+export TERRAFORM_DIR="/opt/terraform/"
+export TERRAFORM_VERSION="0.12.24"
+
+echo "**** Startup Step 4/9: Create tahe directory for Terraform in /opt. ****"
```
bharathkkb, line 13:
```suggestion
echo "**** Startup Step 4/9: Create the directory for Terraform in /opt. ****"
```

0-bootstrap/modules/jenkins/scripts/jenkins_gce_startup_script.sh
```diff
@@ -0,0 +1,31 @@
...
+mkdir $TERRAFORM_DIR && cd $TERRAFORM_DIR || exit
```
bharathkkb, line 14: I am not sure why we have exit here? A startup script would be running as root  and would have required permissions?

0-bootstrap/modules/jenkins/scripts/jenkins_gce_startup_script.sh
```diff
@@ -0,0 +1,31 @@
...
+
+echo "**** Startup Step 7/9: Unzip the Terraform download. ****"
+unzip "terraform_{$TERRAFORM_VERSION}_linux_amd64.zip"
+chmod 755 $TERRAFORM_DIR/terraform
+
+echo "**** Startup Step 8/9: Download and install the Terraform validator ****"
+gsutil cp gs://terraform-validator/releases/2019-04-04/terraform-validator-linux-amd64 .
+chmod 755 $TERRAFORM_DIR/terraform-validator-linux-amd64
+
+echo "**** Startup Step 9/9: Set the Linux PATH to point to the Terraform directory. ****"
```
bharathkkb, line 30: Shouldn't we be also creating a jenkins user?

0-bootstrap/modules/jenkins/scripts/jenkins_gce_startup_script.sh
caleonardo: Done. Thank you.

0-bootstrap/modules/jenkins/scripts/jenkins_gce_startup_script.sh
caleonardo: It shows a lint error. If the mkdir doesn't work for any reason, the script is cancelled

0-bootstrap/modules/jenkins/scripts/jenkins_gce_startup_script.sh
caleonardo: Done.

0-bootstrap/modules/jenkins/scripts/jenkins_gce_startup_script.sh
caleonardo: Yes! Thank you - I created a template for this now.

0-bootstrap/modules/jenkins/scripts/jenkins_gce_startup_script.sh
caleonardo: The user is created automatically when you use the ssh metadata, because the the format of the public key is `"ssh-rsa [KEY_VALUE] [USERNAME]"` - So GCE creates a `jenkins` user automatically

0-bootstrap/modules/jenkins/scripts/jenkins_gce_startup_script.sh
bharathkkb: even with `mkdir -p`?

caleonardo: Actually, the lint message is related to the `cd` command - see https://github.com/koalaman/shellcheck/wiki/SC2164

So I understand we should leave it there.

bharathkkb: `set -e` might be cleaner as it makes sense to exist if any of these steps fail not just the `cd` step.

.gitignore
```diff
@@ -26,7 +26,6 @@ override.tf.json
 *_override.tf
 *_override.tf.json
 .idea/
-.DS_Store
```
rjerrems, line 29: is there a reason this being removed?

0-bootstrap/modules/jenkins/files/jenkins_gce_startup_script.sh
```diff
@@ -0,0 +1,43 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+#!/bin/sh
+
+echo "**** Startup Step 1/9: Update apt-get repositories. ****"
```
rjerrems, line 18: nit: 1/9 when there is only 8 steps

0-bootstrap/modules/jenkins/files/jenkins_gce_startup_script.sh
```diff
@@ -0,0 +1,43 @@
...
+sudo apt-get update
+
+echo "**** Startup Step 2/8: Install Java. ****"
+sudo apt-get install -y default-jdk
+
+echo "**** Startup Step 3/8: Create the directory for Terraform in /opt. ****"
```
rjerrems, line 24: if we are installing the dependencies locally, we should align with the bootstrap / dependencies installed in the container in the boostrap module https://github.com/terraform-google-modules/terraform-google-bootstrap/blob/master/modules/cloudbuild/cloudbuild_builder/Dockerfile

0-bootstrap/modules/jenkins/main.tf
```diff
@@ -112,7 +124,131 @@ resource "google_compute_firewall" "allow_ssh_to_jenkins_agent_fw" {
 
 resource "google_compute_network" "jenkins_agents" {
   project = module.cicd_project.project_id
-  name    = "jenkins-agents-network"
+  name    = "vpc-b-jenkinsagents"
```
rjerrems, line 127: What does the b stand for here? Should this be p for production?

0-bootstrap/modules/jenkins/files/jenkins_gce_startup_script.sh
rjerrems: The broader question is, if we are using containers for builds why do we need these dependencies on the VM?

bharathkkb:
@rjerrems no we will be using a one regular GCE agent so not containers. All the necessary binaries will be installed in this GCE agent.

0-bootstrap/modules/jenkins/files/jenkins_gce_startup_script.sh
rjerrems: There are still dependencies in that docker file which we should include such as jq & alpha / beta components for gcloud. It would be nice to also verify the shasum of the terraform binary given you are passing that information in

caleonardo: I can just implement the same apt-get we have in the Dockerfile:
```
apt-get update && \
   /builder/google-cloud-sdk/bin/gcloud -q components install alpha beta && \
    apt-get -y install curl jq unzip ca-certificates && \
    curl https://releases.hashicorp.com/terraform/${ENV_TERRAFORM_VERSION}/terraform_${ENV_TERRAFORM_VERSION}_linux_amd64.zip \
      > terraform_linux_amd64.zip && \
    echo "${ENV_TERRAFORM_VERSION_SHA256SUM} terraform_linux_amd64.zip" > terraform_SHA256SUMS && \
    sha256sum -c terraform_SHA256SUMS --status && \
    unzip terraform_linux_amd64.zip -d /builder/terraform && \
    rm -f terraform_linux_amd64.zip && \
    apt-get remove --purge -y curl unzip && \
    apt-get --purge -y autoremove && \
    apt-get clean && \
    rm -rf /var/lib/apt/lists/*
```
But, we also have the option of just using the Dockerfile for the pipeline... I am thinking about this because I know you are also working in a standard way of running the pipeline (using the bash scripts).

So, is the greater good to use the Dockerfile? or to install all these binaries in the GCE instance? - If we use the Dockerfile, then we just need to install docker in the GCE instance.

Please let me know what do you think and how to proceed.

Cheers.

caleonardo: @rjerrems and @bharathkkb @morgante -
I just discovered an issue with the installation on packages: The Agent cannot download any package if it doesn't have an external IP. We are not adding an external IP in the final version of the code, because all communication will happen with on-prem and thru the VPN.

So, the solution is to use the container optimised OS (instead of Debian) and use docker in the pipeline. This way we don't have to install Docker either.

The only problem I see is... How will the docker image be built without internet access? - Using private repositories?

rjerrems: Ultimately the agent is going to need internet access in some form, potentially through Cloud NAT. Even if it can reach the APIs privately, it is still going to need to reach Github / Terraform registry to pull the CFT modules as a part of the build process so we will need to solve this problem, unless spend time investigating options to cache the modules someone that is reachable without internet access

0-bootstrap/modules/jenkins/main.tf
```diff
@@ -112,7 +124,131 @@ resource "google_compute_firewall" "allow_ssh_to_jenkins_agent_fw" {
...
+ *****************************************/
+
+resource "google_kms_crypto_key_iam_binding" "jenkins_crypto_key_encrypter" {
+  crypto_key_id = google_kms_crypto_key.tf_key.self_link
+  role          = "roles/cloudkms.cryptoKeyEncrypter"
+
+  members = [
+    "group:${var.group_org_admins}",
+  ]
+}
```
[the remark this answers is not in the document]
rjerrems: in the other module, this is to provide support for people to use KMS with Cloud Build rather than implementing it for a specific use case. You can probably safely remove it from the Jenkins module though if you would prefer to keep it lean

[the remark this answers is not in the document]
caleonardo: Sounds good. I am removing it to avoid confusion with the KSM configuration in the secrets project.

0-bootstrap/modules/jenkins/files/jenkins_gce_startup_script.sh
caleonardo: Thank you @rjerrems!

The team agreed on using Cloud NAT for now and I've implemented that as the solution for this issue. We have in backlog to do something more sophisticated, such as getting binaries and packages from on-prem (thru the VPN) or to filter the websites the Agent will be able to reach thru the NAT.

rjerrems: No worries -  I have unresolved this one, there are still a couple of things missing that are mentioned above. Most importantly:

- jq (used in a number of the modules)
- beta / alpha commands for gcloud

caleonardo: Yes, doing that right now actually. Will send a push later todays with it.

caleonardo:
@rjerrems / @bharathkkb - Can you please take a look at why my tests are failing?

The cloud build tests show an old version of my `main.tf` file:

```
Running terraform fmt
0-bootstrap/modules/jenkins-agent/main.tf
--- old/0-bootstrap/modules/jenkins-agent/main.tf
+++ new/0-bootstrap/modules/jenkins-agent/main.tf
@@ -55,8 +55,8 @@
   // not needed  if user has a golden image with all necessary packages.
   template = file("${path.module}/files/jenkins_gce_startup_script.sh")
   vars = {
-    tpl_TERRAFORM_DIR     = "/opt/builder/terraform/"
-    tpl_TERRAFORM_VERSION = var.terraform_version
+    tpl_TERRAFORM_DIR               = "/opt/builder/terraform/"
+    tpl_TERRAFORM_VERSION           = var.terraform_version
     tpl_TERRAFORM_VERSION_SHA256SUM = var.terraform_version_sha256sum
   }
 }
Error: terraform fmt failed with exit code 3
Check the output for diffs and correct using terraform fmt <dir>
```

But if you look at the commit being referenced (`8dca4f1` - https://github.com/terraform-google-modules/terraform-example-foundation/blob/8dca4f10e0daa8ac7b24873fd727aacc954a8d31/0-bootstrap/modules/jenkins/main.tf#L53), the file doesn't have any issues.

0-bootstrap/modules/jenkins-agent/main.tf
```diff
@@ -0,0 +1,260 @@
...
+  account_id   = format("%s-%s", var.service_account_prefix, var.jenkins_agent_sa_email)
+  display_name = "Jenkins Agent (GCE instance) custom Service Account"
+}
+
+data "template_file" "jenkins_agent_gce_startup_script" {
+  // Add Cloud NAT for the Agent to reach internet and download updates and necessary binaries
+  // not needed  if user has a golden image with all necessary packages.
+  template = file("${path.module}/files/jenkins_gce_startup_script.sh")
+  vars = {
+    tpl_TERRAFORM_DIR     = "/opt/builder/terraform/"
```
bharathkkb, line 58: nit:fmt

bharathkkb: @caleonardo looks like this PR still has it.  8dca4f1 doesn't seem to have any changes for `0-bootstrap/modules/jenkins-agent/main.tf` introduced in 2e1eddd I think.

caleonardo: Found the offending file! - I somehow duplicated a folder. It is fixed now.

Thank you!

caleonardo:
@bharathkkb and @rjerrems - I am considering this absolutely completed. Let me know if you see anything that need my attention.

When you approve it, I will send another PR to activate this module by default and comment the cloudbuild module in `0-bootstrap/main`

Cheers,

bharathkkb:
@caleonardo I maybe mistaken but I thought we deferred on activating Jenkins by default due to the VPN requiring manual steps? 
/cc @rjerrems

0-bootstrap/modules/jenkins-agent/variables.tf
```diff
@@ -86,22 +127,23 @@ variable "project_prefix" {
...
+    "bigquery.googleapis.com", // required to create the BQ log sinks
     "cloudresourcemanager.googleapis.com",
     "cloudbilling.googleapis.com",
     "iam.googleapis.com",
     "admin.googleapis.com",
     "appengine.googleapis.com",
     "storage-api.googleapis.com",
-    "cloudkms.googleapis.com"
+    "cloudkms.googleapis.com",
+    "sourcerepo.googleapis.com",
```
bharathkkb, line 146: do we need source repos? My assumption was the Jenkins is targeting users with other git hosting solutions anyway.

0-bootstrap/modules/jenkins-agent/README.md
```diff
@@ -46,54 +48,56 @@ Error: google: could not find default credentials. See https://developers.google
...
-
-
-## Resources created
-
-- TODO
-    - KMS Keyring and key for secrets, including IAM for Cloudbuild, Org Admins and Terraform service account
-    - (optional) Cloudbuild impersonation permissions for a service account
-    - (optional) Cloud Source Repos, with triggers for terraform plan (all other branches) & terraform apply (master)
-
+    1.  Add Cloud NAT for the Agent to reach internet and download updates and necessary binaries (not needed if user has a golden image with all necessary packages)
```
bharathkkb, line 58: Is NAT still a TODO?

rjerrems:
> @caleonardo I maybe mistaken but I thought we deferred on activating Jenkins by default due to the VPN requiring manual steps?
> /cc @rjerrems

Yep that is my understanding, what we have agreed is that Jenkins will have clear instructions how to turn on / use - but default will be Cloud Build so that the example is usable without manual config.

rjerrems:
FYI I have implemented @bharathkkb 's feedback as well as a couple of other optimizations for the bootstrap script and to ensure that binaries are in the path for non privledged users. If you are ok with them @caleonardo we can probably merge.

0-bootstrap/modules/jenkins-agent/variables.tf
caleonardo: Good catch. removing this and kms as well.

caleonardo:
> > @caleonardo I maybe mistaken but I thought we deferred on activating Jenkins by default due to the VPN requiring manual steps?
> > /cc @rjerrems
> 
> Yep that is my understanding, what we have agreed is that Jenkins will have clear instructions how to turn on / use - but default will be Cloud Build so that the example is usable without manual config.

kk, Will send the last PR adding Jenkins commented. That will be a separate PR anyways.

rjerrems:
@caleonardo - did you mean to clobber the changes I added? I have re-instated them locally and can push if required. The main things are:
- exposing skip_glcoud_download
- Updating project factory to version 8
- Changing binary path to `/usr/local/bin/`
- Removing sudo from metadata script as it runs as root

PR #86, opened 2020-07-16
refactor: folder structure for 1-org

rjerrems:
To align with the new environment model / code structure


daniel-cit:
LGTM.

Only one nit and two questions regrading the `README.md` file:

1) In the manual instructions section we have:

```
### Run terraform locally
1. Change into 1-org folder
```

but now it should be

```
### Run terraform locally
1. Change into 1-org/envs/shared/ folder
```

2) In the Cloud build part, should the branch name change from *plan* to *shared* like for example from `git checkout -b plan` to  `git checkout -b shared` ?

3) In the Cloud build part, should the full relative path be added to the *backend* and *tfvar* files, like for example from `terraform.example.tfvars` to `envs/shared/terraform.example.tfvars` ?

rjerrems:
Thanks for the review and feedback @daniel-cit .

1. Will update the docs

2. As per comments from @morgante - I think the plan in terms of branching is that `shared` will be applied on the `prod` branch. We will need to add some documentation to make this clear as it departs somewhat from the direct branch -> environment mapping.

3. I think the wrapper script should handle this behaviour, which @mikelaramie is working on :)

bharathkkb:
+1 I think overall we need some additional docs around our branching strategy. Added #87 to track this.

PR #89, opened 2020-07-17
flattened common folder, updated project names

mikelaramie:
removed subfolders from 'common', updated project names to include 'prj-' prefix


mikelaramie:
Looks like I forgot to update the folder data objects in 2-environments.  I'll clean it up and resubmit.

1-org/projects.tf
```diff
@@ -150,6 +150,6 @@ module "dns_hub" {
 
   labels = {
     environment      = "prod"
-    application_name = "prj-dns-hub"
+    application_name = "dns-hub"
```
daniel-cit, line 153: if the label is fixed, the search for the label needs to be fixed too: 
https://github.com/terraform-google-modules/terraform-example-foundation/blob/e7ef9c22071fd562aba400ff0026cfb60a738a30/3-networks/envs/shared/main.tf#L26

1-org/projects.tf
```diff
@@ -97,7 +97,7 @@ module "interconnect" {
 
   labels = {
     environment      = "prod"
-    application_name = "prj-interconnect"
+    application_name = "interconnect"
```
daniel-cit, line 100: if the label is fixed, the search for the label needs to be fixed too: 
https://github.com/terraform-google-modules/terraform-example-foundation/blob/e7ef9c22071fd562aba400ff0026cfb60a738a30/3-networks/modules/dedicated_interconnect/main.tf#L22

1-org/projects.tf
mikelaramie: Good catch; I'll fix those as well

rjerrems:
Hey @mikelaramie - it looks like there are a bunch of changes showing up on this PR and your other one that aren't a part of the changes you made but are from previous commits to develop?

mikelaramie:
> Hey @mikelaramie - it looks like there are a bunch of changes showing up on this PR and your other one that aren't a part of the changes you made but are from previous commits to develop?

@rjerrems I think I rebase'd incorrectly.  I'll close this one and submit a cleaner PR.

PR #91, opened 2020-07-17
added wrapper script, updated CB yamls

mikelaramie:
Added tf-wrapper.sh and updated the Cloud Build yamls to use the wrapper script.  

TODO:  Update READMEs to include copying the wrapper script.


mikelaramie:
Sorry about leaving out the READMEs; need to step away for a bit and wanted to get this checked in in case anyone needed to test it.  I'll try and do a PR for the READMEs later this afternoon, or feel free to reject this PR if we want to wait.

build/tf-wrapper.sh
```diff
@@ -0,0 +1,101 @@
+##TODO:  better error handling
+##TODO:  variable for path?
+##TODO:  How does this work for monorepo?
+##TODO:  
+
+#!/bin/bash
+branchname=$2
+policyrepo=$3
+
+tf_validator_path="./terraform-validator-linux-amd64"
```
bharathkkb, line 10: @rjerrems I see that we have this dockerfile https://github.com/terraform-google-modules/terraform-google-bootstrap/blob/master/modules/cloudbuild/cloudbuild_builder/Dockerfile that is built, can we add tf-validator bin in there?

rjerrems: Yeah agree we should include that as a part of that container that gets built

build/tf-wrapper.sh
```diff
@@ -0,0 +1,101 @@
...
+tf_plan() {
+  terraform plan -input=false -out $branchname.tfplan || exit 21
+}
+
+tf_validate() {
+  if ! ${tf_validator_path} version &> /dev/null; then
+    terraform validate || exit 31
+  else
+    terraform show -json $branchname.tfplan > $branchname.json || exit 32
+    terraform-validator-linux-amd64 validate $branchname.json --policy-path=${policyrepo} || exit 33
```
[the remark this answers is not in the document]
rjerrems: I'm not 100% git is available in the container was well, so will need to check that if this is the case

build/tf-wrapper.sh
caleonardo: qq: Should this wrapper call this script that finds the credentials?: https://github.com/terraform-google-modules/terraform-google-bootstrap/blob/master/modules/cloudbuild/cloudbuild_builder/entrypoint.bash

There is a point at which the cicd tool (either Jenkins or Cloud Build) has to impersonate the terraform Service account. I am not sure exactly when this is happening in this script

PR #94, opened 2020-07-18
add missing sinks and enable data access logs.

daniel-cit:
Changes in this PR:

- add logging sinks for: "VPC Flow Logs", "Firewall Rules Logging" and "Access Transparency logs".
- add "google_organization_iam_audit_config" or "google_folder_iam_audit_config" with toggle(default = true) for "allServices" with DATA_READ, DATA_WRITE and ADMIN_READ.
- update README with disclaimer regarding the cost of data access logs.

@bharathkkb @rjerrems @morgante Could you please take a look at this change?

### DIFF
--- a/1-org/envs/shared/log_sinks.tf
+++ b/1-org/envs/shared/log_sinks.tf
@@ -73,7 +73,7 @@ module "bq_system_event_logs" {
 *****************************************/
 
 resource "google_organization_iam_audit_config" "organization_data_access_config" {
-  count   = var.data_access_logs_enabled ? (var.parent_folder != "" ? 0 : 1) : 0
+  count   = var.data_access_logs_enabled && var.parent_folder == "" ? 1 : 0
   org_id  = var.org_id
   service = "allServices"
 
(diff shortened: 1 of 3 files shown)


1-org/envs/shared/log_sinks.tf
```diff
@@ -72,6 +72,42 @@ module "bq_system_event_logs" {
   Audit Logs - Data Access
 *****************************************/
 
+resource "google_organization_iam_audit_config" "organization_data_access_config" {
+  count   = var.data_access_logs_enabled ? (var.parent_folder != "" ? 0 : 1) : 0
```
bharathkkb, line 76: cant we just do ` var.data_access_logs_enabled && var.parent_folder == ""  1 : 0 `
If I understand the logic correctly, if data_access_logs_enabled and if org i.e `parent_folder == "" ` then we need one `google_organization_iam_audit_config`

daniel-cit: You are right. Thanks for the suggestion.

PR #102, opened 2020-07-23
Fix google provider version in 0-bootstrap step

daniel-cit:
Fixes `Error: no suitable version is available`  in  0-bootstrap step

```
No provider "google" plugins meet the constraint " < 4.0, < 4.0, < 4.0, < 4.0, < 4.0, < 4.0,>= 2.1,>= 2.1,>= 2.1,>= 2.1,>= 3.8,>= 3.8,~> 3.30.0,~> 3.31,~> 3.5".

```

@rjerrems @morgante @bharathkkb  could you please check this fix?

### DIFF
--- a/0-bootstrap/main.tf
+++ b/0-bootstrap/main.tf
@@ -15,11 +15,11 @@
  */
 
 provider "google" {
-  version = "~> 3.30.0"
+  version = "~> 3.30"
 }
 
 provider "google-beta" {
-  version = "~> 3.30.0"
+  version = "~> 3.30"
 }
 
 provider "null" {


bharathkkb:
@rjerrems did you mean to limit the bootstrap-module to `>= 3.31 <4`? 
https://github.com/terraform-google-modules/terraform-google-bootstrap/blob/master/versions.tf

rjerrems:
I was just aligning the provider versions to latest wherever it was referenced. I supposed in versions file, we can probably relax that somewhat to `>= 3.9` where the fix for default networks deletion was introduced

bharathkkb:
@rjerrems gotcha, yeah I think we dont need to bump the lower bound unless a new feature in the provider being used in the module. Usually we version lock when we introduce a new feature to a module from the latest provider and we want to prevent folks who are at a lower provider version from accidentally using it and getting an error.

PR #100, opened 2020-07-22
Adding tf-wrapper

mikelaramie:
Adding the tf-wrapper script

TODO:  Update documentation to include tf-wrapper.sh in the `cp cloudbuild-*` (or, should we consider naming it cloudbuild-tf-wrapper.sh?)

PR #106, opened 2020-07-23
refactor: rename backends, add steps to automate updating

rjerrems:
This change makes sure all sample backends are named `backend.tf.example` and adds commands that can be used to rename and update them with the bucket that is created in the bootstrap step.


morgante:
I thought we were switching to an approach where manually renaming all buckets is *not* required. I think it's onerous and would like to instead use the sed-based approach (where we automatically rename the backends as part of the pipeline).

@bharathkkb @mikelaramie Has this changed?

rjerrems:
I don't mind either way, but this approach also is not overly onerous. Its just run once at the beginning and has the added benefit of the committed code being an accurate reflection of the backend being used.

rjerrems:
Also FYI - we currently have a mix of `backend.tf` and `backend.tf.example`

bharathkkb:
@morgante I believe this change is particularly for those who would try this locally without a CI system. I was not aware that we were planning to put this in the pipeline, but if we were, I assume it would be a simple change i.e if backend.tf.example found change to backend.tf and update the bucketname with a parameter/substitution using sed.
/cc @mikelaramie

morgante:
> One question I have though, if we are going to dynamically use sed to replace the bucket - it doesn't seem all that more safe / different from supplying it on the CLI. With something like:

It's much safer.

If I run `terraform apply` locally without replacing the backend it will error out and not apply until I can change.

If we omit backend.tf and I accidentally run `terraform apply` without injecting the backend at the CLI, Terraform will assume I want local state and attempt to recreate all my infra.

mikelaramie:
This should be easy enough to add in since we're already passing _STATE_BUCKET_NAME in the build triggers.

    sed -i "s/UPDATE_ME/${_STATE_BUCKET_NAME/g" backend.tf

rjerrems:
@morgante @mikelaramie  I have:
- Updated files from `backend.tf.example` to `backend.tf` with the exception of bootstrap, which is designed to be executed locally and renamed after initial execution.
- Added a step to cloudbuild yamls to replace the bucket in the files
- Updated docs to have users run the command if they are running locally

build/cloudbuild-tf-apply.yaml
```diff
@@ -14,12 +14,16 @@
...
+- id: 'setup'
+  name: gcr.io/$PROJECT_ID/terraform
   entrypoint: /bin/bash
   args:
   - -c
   - |
+    echo "Setting up gcloud for impersonation"
     gcloud config set auth/impersonate_service_account ${_TF_SA_EMAIL}
+    echo "Adding bucket information to backends"
+    for i in `find -name 'backend.tf'`; do sed -i 's/UPDATE_ME/${_STATE_BUCKET_NAME}/' $i; done
```
daniel-cit, line 26: will this change conflict with the change `sed -i "s/UPDATE_ME/${_STATE_BUCKET_NAME}/g" ./envs/*/backend.tf` in PR https://github.com/terraform-google-modules/terraform-example-foundation/pull/100

morgante: Agreed, we can probably consolidate this logic in the update script.

mikelaramie: i like the `find` method better.  I think at this point almost anything in PR #100 is deprecated

PR #108, opened 2020-07-23
Adds integration test for step 0-bootstrap

atos-cit:
This pull request adds integration tests for step 0-bootstrap, please take a look.

### DIFF
--- a/test/integration/bootstrap/controls/cloudbuild.rb
+++ b/test/integration/bootstrap/controls/cloudbuild.rb
@@ -1,0 +1,46 @@
+# Copyright 2020 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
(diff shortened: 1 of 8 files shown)


test/integration/bootstrap/controls/cloudbuild.rb
```diff
@@ -0,0 +1,46 @@
...
+    it { should exist }
+  end
+
+  csr_repos.each do |repo, _|
+    describe google_sourcerepo_repository(project: cloudbuild_project_id, name: repo) do
+      it { should exist }
+    end
+  end
+
+  google_cloudbuild_triggers(project: cloudbuild_project_id).ids.each do |id|
```
amandakarina, line 39: This test case doesn't check if all requested triggers were created, only that all created triggers have at least one of the branch_name and repo_name. We can use the gcloud to check this correctly.

PR #109, opened 2020-07-23
Jenkins agent jenkinsfile 5 - missing scopes and roles

caleonardo:
Added missing scopes and roles and updated terraform validator version


0-bootstrap/modules/jenkins-agent/files/jenkins_gce_startup_script.sh
```diff
@@ -42,7 +42,7 @@ wget "https://releases.hashicorp.com/terraform/${tpl_TERRAFORM_VERSION}/terrafor
     rm -rf /var/lib/apt/lists/*
 
 echo "**** Startup Step 6/9: Download and install the Terraform validator ****"
-gsutil cp gs://terraform-validator/releases/2019-04-04/terraform-validator-linux-amd64 .
+gsutil cp gs://terraform-validator/releases/2020-03-05/terraform-validator-linux-amd64 .
```
bharathkkb, line 45: I think #107 is also updating this.

build/Jenkinsfile
```diff
@@ -0,0 +1,35 @@
+pipeline {
```
bharathkkb, line 1: Do we have any instruction w.r.t how users should be using this Jenkinsfile?

0-bootstrap/modules/jenkins-agent/files/jenkins_gce_startup_script.sh
caleonardo: Yes, you are right. I will leave the change here anyways. Thank you.

build/Jenkinsfile
caleonardo: Yes, but it can be much clearer. I will add a note about the need of copying `Jenkinsfile` and `tf-wrapper.sh` to the root folders or the new repos (when people distribute the mono-repo to multi-repo)

0-bootstrap/modules/jenkins-agent/main.tf
```diff
@@ -95,10 +95,11 @@ resource "google_compute_instance" "jenkins_agent_gce_instance" {
   service_account {
     email = google_service_account.jenkins_agent_gce_sa.email
     scopes = [
-      // TODO(caleonardo): These scopes will need to change.
+      // TODO(caleonardo): These scopes should be reduced and avoid using "cloud-platform".
       "https://www.googleapis.com/auth/compute.readonly",
-      "https://www.googleapis.com/auth/devstorage.read_only",
+      "https://www.googleapis.com/auth/devstorage.read_write",
```
rjerrems, line 100: If you have cloud-platform, you probably don't need this scope anymore. Also, given the number of APIs this service account will be hitting, it may be an ongoing game of whack-a-mole trying to reduce this in future. I agree with @mikelaramie 's sentiment around focusing on the IAM permissions in this case

caleonardo: Thank you @rjerrems - If the GCE Agent will impersonate the Terraform seed SA to make the changes, can the Agent have reduced scopes? - if so, I think we should reduce them.  

If the Agent needs all the scopes anyways (I mean, even using the impersonation), then yes, we can just use `cloud-platform`.

rjerrems: I think we will need to test this, but my assumption would be that even with impersonation all of the scopes would be required. Happy to reduce if we can

mikelaramie: It's in our [best practices](https://cloud.google.com/compute/docs/access/create-enable-service-accounts-for-instances#best_practices) to allow `cloud-platform` and rely on IAM to restrict what a Service Account can or can't do.  Scopes can provide a false sense of security, because they only apply to the GCE instance making the API call.

caleonardo: Thank you for the details @mikelaramie - This is now only using scope `cloud-platform`.

0-bootstrap/README.md
```diff
@@ -22,6 +22,25 @@ Further details of permissions required and resources created, can be found in t
 1. Copy the backend by running `cp backend.tf.example backend.tf` and update `backend.tf` with your bucket from the apply step (The value from `terraform output gcs_bucket_tfstate`)
 1. Re-run `terraform init` agree to copy state to gcs when prompted
     1. (Optional) Run `terraform apply` to verify state is configured correctly
+
+### If you are using Jenkins
+
+If you are using the `jenkins_bootstrap` sub-module, please see the [README](./modules/jenkins-agent/README.md) for the requirements and instructions on how to run the bootstrap step, which include implementing VPN, configuring your Jenkins Master among other steps.
+
+### Update all example backends with your bucket information
```
rjerrems, line 30: This is out of date with latest code. There should only be one file called `backend.tf.example`, the rest are all `backend.tf` now. 

See latest docs on master https://raw.githubusercontent.com/terraform-google-modules/terraform-example-foundation/develop/0-bootstrap/README.md 

These steps are now optional, for when you are running terraform locally.

0-bootstrap/modules/jenkins-agent/README.md
```diff
@@ -188,69 +191,104 @@ You will need this key pair to enable authentication between the Master and Agen
...
-  - VPN details from your Network administrator:
-     - VPN public IP
-     - VPN PSK secret
-     - Private IP address for the Jenkins Agent (which will be created in the `cicd` GCP Project in Step "2.4. Run terraform commands"). This private IP will be reachable through the VPN connection that you will create in Step "3.2. Configure VPN Network tunnel".
+  - Public SSH key generated in the Jenkins Master (this is not a secret and can be in your repository code). Show the public key using `cat "${SSH_KEY_FILE_PATH}.pub"`, you will have to copy / paste it in the the `terraform.tfvars` file.
 
-**2.1. Clone the Repository provided here**
+Although this infrastructure code is distributed to you as a [monorepo](https://github.com/terraform-google-modules/terraform-example-foundation), you will store it in five different repositories, one for each directory (`0-bootstrap, 1-org, 2-environments, 3-networks, 4-projects`). Here we will work with the first of these repositories <YOUR_NEW_REPO_CLONE-0-bootstrap>.
 
+- 1.3.1. Clone this repository:
```
rjerrems, line 208: We usually try to avoid numbering that is not incremented automatically.

build/Jenkinsfile
```diff
@@ -0,0 +1,35 @@
...
+  agent {
+      label "gcp-agent"
+  }
+  stages {
+    stage('TF init') {
```
bharathkkb, line 6: Add an additional stage for parity with CB steps.
Ref: https://github.com/terraform-google-modules/terraform-example-foundation/pull/100#discussion_r460332673

0-bootstrap/modules/jenkins-agent/README.md
caleonardo: Done. Thank you.

0-bootstrap/modules/jenkins-agent/README.md
```diff
@@ -116,8 +116,8 @@ Error: failed pre-requisites: missing permission on "billingAccounts/aaaaaa-bbbb
 
 -   [gcloud sdk](https://cloud.google.com/sdk/install) >= 206.0.0
 -   [Terraform](https://www.terraform.io/downloads.html) >= 0.12.6
--   [terraform-provider-google] plugin 2.1.x
--   [terraform-provider-google-beta] plugin 2.1.x
+-   [terraform-provider-google](https://github.com/terraform-providers/terraform-provider-google) plugin 2.1.x
```
bharathkkb, line 119: Not sure why we are calling out `plugin 2.1.x.

build/Jenkinsfile
```diff
@@ -0,0 +1,43 @@
+pipeline {
+  agent {
+      label "gcp-agent"
+  }
+  stages {
+    stage('setup') {
+      steps {
+          sh '''
+          ./tf-wrapper.sh setup $BRANCH_NAME
```
bharathkkb, line 9: need to align with the impersonate step as discussed offline

0-bootstrap/modules/jenkins-agent/README.md
caleonardo: I originally copied this from the bootstrap module: https://github.com/terraform-google-modules/terraform-google-bootstrap/blob/master/README.md#software

But I can remove it now in my next push

build/Jenkinsfile
caleonardo: Jenkinsfile now has the extra step and it is using the same command as the `cloud-build-tf-plan.yaml` and `cloud-build-tf-apply.yaml` files:

```
    echo "Setting up gcloud for impersonation"
    gcloud config set auth/impersonate_service_account ${_TF_SA_EMAIL}
    echo "Adding bucket information to backends"
    for i in `find -name 'backend.tf'`; do sed -i 's/UPDATE_ME/${_STATE_BUCKET_NAME}/' $i; done
```

0-bootstrap/README.md
```diff
@@ -62,13 +66,8 @@ Currently, the bucket information is replaced in the state backends as a part of
...
-## Instructions
-
-If you are using `jenkins_bootstrap`, please follow the instructions on how to run the bootstrap step with the `jenkins_bootstrap` sub-module described in the [README](./modules/jenkins-agent/README.md), which include implementing VPN, configuring your Jenkins Master among other steps.
-
 ### Software
 
--   [gcloud sdk](https://cloud.google.com/sdk/install) >= 206.0.0
--   [Terraform](https://www.terraform.io/downloads.html) >= 0.12.6
+- [gcloud sdk](https://cloud.google.com/sdk/install) >= 206.0.0
+- [Terraform](https://www.terraform.io/downloads.html) = 0.12.24
```
bharathkkb, line 72: I am not sure if we should be enforcing `= 0.12.24`, we could keep it the same as before with the warning below if needed.

1-org/envs/shared/README.md
```diff
@@ -29,7 +29,51 @@ You can choose not to enable the Data Access logs by setting variable `data_acce
...
+    ```
+1. Ensure wrapper script can be executed `chmod 755 ./tf-wrapper.sh`.
+1. Rename `terraform.example.tfvars` to `terraform.tfvars` and update the file with values from your environment. (you can re-run `terraform output` in the 0-bootstrap directory to find these values). Make sure that `default_region` is set to a valid [BigQuery dataset region](https://cloud.google.com/bigquery/docs/locations).
+    ```
+    # Rename file
+    mv envs/shared/terraform.example.tfvars envs/shared/terraform.tfvars
+    # Edit the file to provide the necessary values
+    vi envs/shared/terraform.tfvars
+    ```
+1. Run this command to update the `backend.tf` file with the Terraform State GCS bucket name. Replace the `TF_STATE_GCS_BUCKET_NAME` with the name of your bucket (you can re-run `terraform output` in the 0-bootstrap directory to find these values).
```
bharathkkb, line 59: We shouldnt need this step now as Jenkinsfile takes care of this in setup stage?

0-bootstrap/README.md
bharathkkb: changed via f68b382

1-org/envs/shared/README.md
bharathkkb: changed via f68b382

bharathkkb:
@rjerrems I think the team will be doing some Jenkins testing today and would be great to get this in if you get a chance.

PR #111, opened 2020-07-24
Fix readmes for cloud build with wrapper script and add missing symbolic link

### LINKED ISSUE #110: 2-environments/dev is missing symbolic link
2-environments/dev/terraform.tfvars  is missing the symbolic link to  `../../terraform.tfvars`

daniel-cit:
This PR fixes READMEs for all the steps for the execution of the Cloud Build deploy path with the new wrapper script.

It was also added the creation of a missing symbolic link in `2-environments/envs/dev` needed by the instructions. Fixes #110 

Depends on the script and cloud build files from PR https://github.com/terraform-google-modules/terraform-example-foundation/pull/100 and should be merged after PR #100 

@bharathkkb @rjerrems @mikelaramie @morgante Cloud you please take a look at these changes ?

### DIFF
--- a/3-networks/README.md
+++ b/3-networks/README.md
@@ -24,9 +24,14 @@ The purpose of this step is to :
 1. Ensure wrapper script can be executed `chmod 755 ./tf-wrapper.sh`.
 1. Within each envs/ folder rename terraform.example.tfvars to terraform.tfvars and update the file with values from your environment and bootstrap.
 1. Commit changes with `git add .` and `git commit -m 'Your message'`
+1. You will need only once to manually plan + apply the `shared` environment since dev, nonprod and prod depend on it.
+    1. cd to ./envs/shared/
+    1. Update backend.tf with your bucket name from the bootstrap step.
+    1. Run `terraform plan` and review output
+    1. Run `terraform apply`
+    1. If you would like the bucket to be replaced by cloud build at run time, change the bucket name back to `UPDATE_ME
 1. Push your plan branch to trigger a plan `git push --set-upstream origin plan` (the branch `plan` is not a special one. Any branch which name is different from `dev`, `nonprod` or `prod` will trigger a terraform plan).
     1. Review the plan output in your cloud build project https://console.cloud.google.com/cloud-build/builds?project=YOUR_CLOUD_BUILD_PROJECT_ID
-1. Plan will fail for dev/nonprod/prod until prod has been applied at least once since all envs depend on shared (DNS Hub) that is deployed in the prod branch.
 1. Merge changes to prod with `git checkout -b prod` and `git push origin prod`
     1. Review the apply output in your cloud build project https://console.cloud.google.com/cloud-build/builds?project=YOUR_CLOUD_BUILD_PROJECT_ID
 1. After prod has been applied apply dev and nonprod
--- a/3-networks/README.md
+++ b/3-networks/README.md
@@ -10,7 +10,7 @@ The purpose of this step is to :
 1. 0-bootstrap executed successfully.
 1. 1-org executed successfully.
 1. 2-environments executed successfully.
-1. Obtain the value for the access_context_manager_policy_id variable. Can be obtained by running `gcloud access-context-manager policies list --organization YOUR-ORGANIZATION_ID`. It is the value under `name`.
+1. Obtain the value for the access_context_manager_policy_id variable. Can be obtained by running `gcloud access-context-manager policies list --organization YOUR-ORGANIZATION_ID --format="value(name)"`.
 
 ## Usage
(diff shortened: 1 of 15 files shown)


3-networks/README.md
```diff
@@ -9,10 +9,39 @@ The purpose of this step is to:
...
-1. 2-environments/envs/[dev|nonprod|prod] executed successfully.
+1. 2-environments executed successfully.
+1. Obtain the value for the access_context_manager_policy_id variable. Can be obtained by running `gcloud access-context-manager policies list --organization YOUR-ORGANIZATION_ID`. It is the value under `name`.
 
 ## Usage
 
-Follow the instructions in the README.md files of each individual 3-networks/envs/ folder.
+### Setup to run via Cloud Build
+
+1. Clone repo `gcloud source repos clone gcp-networks --project=YOUR_CLOUD_BUILD_PROJECT_ID`
```
bharathkkb, line 19: nit: `YOUR_CLOUD_BUILD_PROJECT_ID` we should call out that this comes from stage `0-bootstrap` if we havent already.

3-networks/envs/prod/README.md
```diff
@@ -8,27 +8,14 @@ The purpose of this step is to setup private and restricted shared VPCs with def
 1. 1-org executed successfully.
 1. 2-environments/envs/prod executed successfully.
 1. 3-networks/envs/shared executed successfully.
+1. Obtain the value for the access_context_manager_policy_id variable. Can be obtained by running `gcloud access-context-manager policies list --organization YOUR-ORGANIZATION_ID`. It is the value under `name`.
```
bharathkkb, line 11: nit: I would just add `--format="value(name)"` so that it is just gives the required value and can be scripted.

3-networks/README.md
```diff
@@ -9,10 +9,39 @@ The purpose of this step is to:
...
+1. Change freshly cloned repo and change to non master branch `git checkout -b plan`
+1. Copy contents of foundation to new repo `cp -R ../terraform-example-foundation/3-networks/* .` (modify accordingly based on your current directory)
+1. Copy cloud build configuration files for terraform `cp ../terraform-example-foundation/build/cloudbuild-tf-* . ` (modify accordingly based on your current directory)
+1. Copy terraform wrapper script `cp ../terraform-example-foundation/build/tf-wrapper.sh . ` (modify accordingly based on your current directory)
+1. Ensure wrapper script can be executed `chmod 755 ./tf-wrapper.sh`.
+1. Within each envs/ folder rename terraform.example.tfvars to terraform.tfvars and update the file with values from your environment and bootstrap.
+1. Commit changes with `git add .` and `git commit -m 'Your message'`
+1. Push your plan branch to trigger a plan `git push --set-upstream origin plan`
```
bharathkkb, line 27: nit: this makes it seem like `plan` is a special branch like `prod,dev,nonprod`. We should say any arbitrary branch will trigger plan.
Additionally @rjerrems I was discussing with @daniel-cit offline if it makes sense to have a small separate section where we explain the branching strategy. i.e we care about `prod,dev,nonprod` and these should ideally be protected, how to merge between these etc. WDYT?

daniel-cit:
Fixes #117 and the part related to copying the script in https://github.com/terraform-google-modules/terraform-example-foundation/issues/115

3-networks/README.md
daniel-cit: step `1-org` has the instruction for getting the correct value:  `(this is from terraform output from the previous section, 0-bootstrap)`.

mikelaramie: possible future feature request:  Would it work well if we labeled the cloud build project during creation so this could be programmatically queried?  Something like:
`gcloud source repos clone gcp-org \
--project=$(gcloud projects list --filter="(labels.application_name:cloudbuild \
AND labels.env:seed)" --format="value(project_id)")`

3-networks/envs/prod/README.md
daniel-cit: Good idea. Thanks

3-networks/README.md
daniel-cit: fixed

3-networks/README.md
daniel-cit: @mikelaramie we could have the this problem too: https://github.com/terraform-google-modules/terraform-example-foundation/issues/113 . Maybe we cloud just help a little more and give the terraform command `terraform output cloudbuild_project_id` in step `1-org`

bharathkkb: I think #113 is just a byproduct of testing. irl, we will only have one seed right? But either approach sgtm.

PR #112, opened 2020-07-25
Add more sinks destinations

daniel-cit:
Adding Pub/Sub topic and  Storage bucket as logging sinks destination and unifying BigQuery destination

Now we have three sinks

- One using BigQuery as destination
- One using Cloud storage as destination
- One using na Pub/Sub topic as destination

All sinks use the same filter for the logs:

```
    logName: /logs/cloudaudit.googleapis.com%2Faccess_transparency
```


rjerrems:
Why are we grouping all of the log sinks together into one filter and export now? Doing this, you loose the ability to set different retention periods for the different log types. In my experience it is pretty common, to have different retention periods as different log types are significantly more verbose than others and often have different requirements for how long they are kept.

It also seems wasteful to send the same logs, to three different destinations by default which will increase the cost and not all of these will be required for every customer. Out of the box you will have the same set of logs in Stackdriver logging, BigQuery, GCS and Pubsub.

rjerrems:
For purposes of keeping PR with added context, the reasoning behind the three locations:

```Just to clarify why we have 3 replica's - the replica's serve 3 different purposes. The GCS bucket is intend for a regulatory/forensic purpose and usually involves bucket lock. The BQ export is for ad hoc analytics by a security team, while the pubsub is used to export to an external siem.```

rjerrems:
I think in subsequent release, we should re-evaluate how this is structured to allow greater flexibility.

konradclapa:
+ for separating this into different modules. We would hardly ever want to deploy all flavors of the sinks.

PR #131, opened 2020-07-27
refactor: add folders and tidy up naming/structure

rjerrems:
This PR adds a business unit folder inside the environment folder for application projects and tidies up the naming.

Right now, projects get created at the top of the hierarchy and there are multiple projects with the same name:

```
org > bu2-p-sample    
org > bu2-p-sample             
org > bu2-p-sample-single
```

This PR changes it to:

```
org > prod > bu2-p-sample-floating
org > prod > bu2-p-sample-private
org > prod > bu2-p-sample-restricted
```

- [x] Refactor business unit 1

cc: @mikelaramie


4-projects/business_unit_2/dev/example_restricted_shared_vpc_project.tf
```diff
@@ -30,26 +30,10 @@ module "restricted_shared_vpc_project" {
   vpc_service_control_perimeter_name = "accessPolicies/${var.policy_id}/servicePerimeters/${var.perimeter_name}"
 
   # Metadata
-  project_prefix    = "bu2-d-sample"
-  application_name  = "bu2-sample-application2"
+  project_prefix    = "bu2-p-sample-restricted"
```
daniel-cit, line 33: This one should be `bu2-d-sample-restricted` instead of `bu2-p-sample-restricted`

4-projects/business_unit_2/dev/example_floating_project.tf
```diff
@@ -14,21 +14,19 @@
...
   environment                 = "dev"
   env_code                    = "d"
-  skip_gcloud_download        = var.skip_gcloud_download
-
-  folder_id = var.parent_folder
 
   # Metadata
-  project_prefix    = "bu2-d-sample-single"
-  application_name  = "bu2-sample-application2"
+  project_prefix    = "bu2-p-sample-floating"
```
daniel-cit, line 28: this one should be  `bu2-d-sample-floating` instead of `bu2-p-sample-floating`

4-projects/business_unit_2/nonprod/example_floating_project.tf
```diff
@@ -16,21 +16,19 @@
...
   environment                 = "nonprod"
   env_code                    = "n"
-  skip_gcloud_download        = var.skip_gcloud_download
-
-  folder_id = var.parent_folder
 
   # Metadata
-  project_prefix    = "bu2-n-sample-single"
-  application_name  = "bu2-sample-application2"
+  project_prefix    = "bu2-p-sample-floating"
```
daniel-cit, line 30: this one should be `bu2-n-sample-floating` instead of  `bu2-p-sample-floating`

4-projects/business_unit_2/nonprod/example_restricted_shared_vpc_project.tf
```diff
@@ -30,26 +30,10 @@ module "restricted_shared_vpc_project" {
   vpc_service_control_perimeter_name = "accessPolicies/${var.policy_id}/servicePerimeters/${var.perimeter_name}"
 
   # Metadata
-  project_prefix    = "bu2-n-sample"
-  application_name  = "bu2-sample-application2"
+  project_prefix    = "bu2-p-sample-restricted"
```
daniel-cit, line 33: this one should be  `bu2-n-sample-restricted` instead of  `bu2-p-sample-restricted`

mikelaramie:
Hey all, while I agree with this change and have championed for it before (#teamFolders), it goes against the design in the upcoming whitepaper.  Should we pull up internally around this change?

4-projects/business_unit_2/nonprod/example_floating_project.tf
mikelaramie: The env code should be a variable to avoid this, no?

daniel-cit: This is in the context of the `nonprod` example so the only allowed value is `p`. Removing variables that had a value that cloud not be changed was a fix in previous PRs.

bharathkkb:
@rjerrems we discussed this offline with @mikelaramie and Sean and currently guidance from the paper is not to have another folder level. Optionally enabling this via flag was suggested. 
By default it would be `org > prod > bu2-p-sample-private` but if flag is `true` we do `org > prod > business-unit-2 > bu2-p-sample-private` WDYT?

4-projects/business_unit_2/dev/example_restricted_shared_vpc_project.tf
rjerrems: good catch, I will fix in next push

rjerrems:
Hey @mikelaramie @daniel-cit @bharathkkb - I have addressed the feedback in most recent commit, have a look when you get a chance and let me know what you think

PR #141, opened 2020-07-28
fix: add missing prj prefixes

rjerrems:
After recreating all of projects in my test org, I found out that we were missing the prefix "prj"

### DIFF
--- a/4-projects/modules/single_project/main.tf
+++ b/4-projects/modules/single_project/main.tf
@@ -21,7 +21,7 @@ module "project" {
   random_project_id           = "true"
   impersonate_service_account = var.impersonate_service_account
   activate_apis               = var.activate_apis
-  name                        = var.project_prefix
+  name                        = "prj-${var.business_code}-${element(split("", var.environment), 0)}-${var.project_prefix}"
   org_id                      = var.org_id
   billing_account             = var.billing_account
   folder_id                   = var.folder_id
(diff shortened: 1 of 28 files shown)


rjerrems:
Yeah I think we probably can, given all the parts of the name are already supplied to the module. Let me have a look

rjerrems:
I have moved into module, take a look when you get a chance @bharathkkb - I ran this locally over projects in my test orgs without any changes required (from previous commit)

4-projects/modules/single_project/main.tf
```diff
@@ -21,7 +21,7 @@ module "project" {
   random_project_id           = "true"
   impersonate_service_account = var.impersonate_service_account
   activate_apis               = var.activate_apis
-  name                        = var.project_prefix
+  name                        = "prj-${var.business_code}-${element(split("", var.environment), 0)}-${var.project_prefix}"
```
daniel-cit, line 24: Instead of using `${element(split("", var.environment), 0)}` to get the env_code here and in other places in the code, we could use the declared variable env_code `${var.env_code}` or remove the variable `env_code` from the `variables.tf` file if it is not needed any more.

```
variable "env_code" {
  description = "A short form of the environment field"
  type        = string
}
```

rjerrems: This should be resolved in latest commit, I think if we can infer that from environment no need to make users add two variables and potentially forget to update one.

PR #148, opened 2020-07-28
make access context manager policy id usage consistent

daniel-cit:
This change makes access context manager policy id usage consistent between 3-networks and 4-projects.


rjerrems:
Looks like this succeeded under build `911cd874-a90a-497b-850f-13d288d08ee0` but did not report back

PR #150, opened 2020-07-29
Misc documentation fixes

daniel-cit:
In this PR:

- Creation of the initial version of the documentation to configure terraform-validator
- Added a note in the main README.md file regarding project creation quota error.

@morgante @rjerrems @bharathkkb Could you please take a look?

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,6 +3,8 @@ This is an example repo showing how the CFT Terraform modules can be composed to
 The supplied structure and code is intended to form a starting point for building your own foundation with pragmatic defaults you can customize to meet your own requirements. Currently, the code leverages Google Cloud Build for deployment of the Terraform from step 1 onwards.
 Cloud Build has been chosen to allow teams to quickly get started without needing to deploy a CI/CD tool, although it is worth noting the code can easily be executed by your preferred tool.
 
+**Note:** when running the examples in this repository, you may receive an error like `Error code 8, message: The project cannot be created because you have exceeded your allotted project quota.` when applying terraform. That means you have reached your [Project creation quota](https://support.google.com/cloud/answer/6330231). In this case you can use this [Request Project Quota Increase](https://support.google.com/code/contact/project_quota_increase) form to request a quota increase. The `terraform_sa_email` created in `0-bootstrap` should also be listed in Email addresses that will be used to create projects in that support form.
+
 ## Overview
 This repo contains several distinct Terraform projects each within their own directory that must be applied separately, but in sequence.
 Each of these Terraform projects are to be layered on top of each other, running in the following order.
--- a/README.md
+++ b/README.md
@@ -3,8 +3,6 @@ This is an example repo showing how the CFT Terraform modules can be composed to
 The supplied structure and code is intended to form a starting point for building your own foundation with pragmatic defaults you can customize to meet your own requirements. Currently, the code leverages Google Cloud Build for deployment of the Terraform from step 1 onwards.
 Cloud Build has been chosen to allow teams to quickly get started without needing to deploy a CI/CD tool, although it is worth noting the code can easily be executed by your preferred tool.
 
-**Note:** when running the examples in this repository, you may receive an error like `Error code 8, message: The project cannot be created because you have exceeded your allotted project quota.` when applying terraform. That means you have reached your [Project creation quota](https://support.google.com/cloud/answer/6330231). In this case you can use this [Request Project Quota Increase](https://support.google.com/code/contact/project_quota_increase) form to request a quota increase. The `terraform_sa_email` created in `0-bootstrap` should also be listed in Email addresses that will be used to create projects in that support form.
-
 ## Overview
 This repo contains several distinct Terraform projects each within their own directory that must be applied separately, but in sequence.
 Each of these Terraform projects are to be layered on top of each other, running in the following order.
--- a/TERRAFORM_VALIDATOR.md
+++ b/TERRAFORM_VALIDATOR.md
@@ -1,0 +1,65 @@
+# terraform-validator configuration
+
+These are instruction to configure [terraform-validator](https://github.com/GoogleCloudPlatform/terraform-validator) in CI/CD for the example foundation.
+
+These instructions consider that you have cloned `terraform-example-foundation` repo locally and had already configured the private repos for `gcp-org`, `gcp-environments`, `gcp-networks` and `gcp-projects`.
+
+## Run terraform-validator with tf-wrapper.sh script manually
+
+The user running terraform-validator must have at least these roles to run terraform-validator:
+
+- **Browser** - `roles/browser`
+- **Security Reviewer** - `roles/iam.securityReviewer`
+
+Installation steps:
+
+1. Install terraform-validator locally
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -3,6 +3,8 @@ This is an example repo showing how the CFT Terraform modules can be composed to
 The supplied structure and code is intended to form a starting point for building your own foundation with pragmatic defaults you can customize to meet your own requirements. Currently, the code leverages Google Cloud Build for deployment of the Terraform from step 1 onwards.
 Cloud Build has been chosen to allow teams to quickly get started without needing to deploy a CI/CD tool, although it is worth noting the code can easily be executed by your preferred tool.
 
+**Note:** when running the examples in this repository, you may receive an error like `Error code 8, message: The project cannot be created because you have exceeded your allotted project quota.` when applying terraform. That means you have reached your [Project creation quota](https://support.google.com/cloud/answer/6330231). In this case you can use this [Request Project Quota Increase](https://support.google.com/code/contact/project_quota_increase) form to request a quota increase. The `terraform_sa_email` created in `0-bootstrap` should also be listed in Email addresses that will be used to create projects in that support form.
```
rjerrems, line 6: I think we should add this to the README for bootstrap, rather than front and center on the main README

TERRAFORM_VALIDATOR.md
```diff
@@ -0,0 +1,65 @@
...
+ERROR: logging before flag.Parse: I0728 23:16:46.430272    3913 validator.go:186] worker 1 starting
+No violations found.
+```
+
+## Run terraform-validator in Cloud Build
+
+Configuration steps:
+
+1. The cloud build service account will need at least these roles to run terraform-validator:
+   - **Browser** - `roles/browser`
```
rjerrems, line 47: Should we add these permissions to the bootstrap code? We can add as something in the example foundation to make sure that validator works.

@morgante - do these permissions like right to you? (they are the product of me and @daniel-cit tinkering until we got things running)

TERRAFORM_VALIDATOR.md
```diff
@@ -0,0 +1,65 @@
...
+      At environment: envs/shared
+      Using policy from: /home/foo/policy-library
+*****************************************************
+ERROR: logging before flag.Parse: I0728 23:16:46.428406    3913 validator.go:177] starting 2 workers
+ERROR: logging before flag.Parse: I0728 23:16:46.428862    3913 validator.go:186] worker 0 starting
+ERROR: logging before flag.Parse: I0728 23:16:46.430272    3913 validator.go:186] worker 1 starting
+No violations found.
+```
+
+## Run terraform-validator in Cloud Build
```
morgante, line 42: We should emphasize this path (put it first).

TERRAFORM_VALIDATOR.md
morgante: Hmm, Validator shouldn't need any *new* permissions beyond what the Cloud Build account already has.

TERRAFORM_VALIDATOR.md
```diff
@@ -0,0 +1,65 @@
...
+
+Configuration steps:
+
+1. The cloud build service account will need at least these roles to run terraform-validator:
+   - **Browser** - `roles/browser`
+   - **Security Reviewer** - `roles/iam.securityReviewer`
+1. Follow the instructions in [configure access for cloud build service account](https://cloud.google.com/cloud-build/docs/securing-builds/configure-access-for-cloud-build-service-account) to grant these roles.
```
morgante, line 49: Why do they need to do this manually? These should be baked into the bootstrap.

TERRAFORM_VALIDATOR.md
```diff
@@ -0,0 +1,65 @@
...
+1. terraform-validator is already installed in the image used in cloud build.
+1. Open `https://console.cloud.google.com/cloud-build/triggers?project=YOUR_CLOUD_BUILD_PROJECT_ID` (this is from terraform output from 0-bootstrap section)
+1. Edit to each one of your Cloud build triggers and add a new Substitution variable with variable `_POLICY_REPO` and value `/workspace/policy-library`.
```
morgante, line 52: This is too much manual work. We should not require manual effort to get validator working.

TERRAFORM_VALIDATOR.md
```diff
@@ -0,0 +1,65 @@
...
+1. Outside of your clone of terraform-example-foundation repo clone the policy repo `git clone https://github.com/forseti-security/policy-library.git`
```
morgante, line 53: We should direct using a bundle: https://github.com/forseti-security/policy-library/blob/master/docs/bundles/forseti-security.md

TERRAFORM_VALIDATOR.md
rjerrems: does it need permissions in the org, to list/get resources / folders / projects etc? By default, cloudbuild will only be able to see things in its own project and we were running into an error where it was throwing errors because it could not enumerate folders/projects.

morgante: We should be running Validator using the same service account that Terraform runs as.

TERRAFORM_VALIDATOR.md
rjerrems: +1

TERRAFORM_VALIDATOR.md
daniel-cit: The problem regarding using the terraform service account is this one (from https://github.com/GoogleCloudPlatform/terraform-validator#Auth)

> terraform-validator commands need to be able to authenticate to Google Cloud APIs. This can be done by generating a credentials.json file 

But we create a folder/organization policy that `Disable service account key creation`

morgante: Terraform Validator runs using Application Default Credentials. That doesn't require downloading service account keys.

rjerrems: Right now, all of the terraform runs using service account impersonation which is what @daniel-cit is alluding too. The cloud build service account, by default only has very limited permissions.

morgante: Got it. I'm still unclear what security is added through that mechanism though and would opt to broaden the Cloud Build account permissions.

Viewer roles should be sufficient.

rjerrems: There are a couple of main benefits:
- Separation of concerns, you need to know about the service account in order to use it. An arbitrary Cloud Build job wont run with those privileges. 
- impersonation can be granted to a super admin group (org admins group in our case), so they can act as the service account for debugging without having to download/distribute a key. (for what is a highly privileged account)

It does introduce some integration challenges, which I think we can weigh up over time against these benefits ^

morgante: That's fair. I remain skeptical of impersonation actually adding real security benefits, but I see the argument. For now we can stick with it and add the necessary viewer roles to the Cloud Build account.

daniel-cit:
Closing this PR:

- note about project creation quota error has been separated in a new PR
- Terraform-validator will be added to the cloud build + bootstrap so that the only manual step will be to copy the policy repo to the step repos (`gcp-org`,...) in the normal cloud build instruction in each README.md

PR #178, opened 2020-08-04
add missing tags to firewall rules.

daniel-cit:
Add missing firewall tags to:

- windows activation
- allow google private api
- allow google restricted api

### DIFF
--- a/3-networks/modules/standard_shared_vpc/firewall.tf
+++ b/3-networks/modules/standard_shared_vpc/firewall.tf
@@ -38,7 +38,7 @@ resource "google_compute_firewall" "deny_all_egress" {
 
 
 resource "google_compute_firewall" "allow_private_api_egress" {
-  name           = "fw-${var.environment_code}-shared-private-65534-e-a-allow-google-api-all-tcp-443"
+  name           = "fw-${var.environment_code}-shared-private-65534-e-a-allow-google-apis-all-tcp-443"
   network        = module.main.network_name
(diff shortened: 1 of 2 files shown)


3-networks/modules/standard_shared_vpc/firewall.tf
```diff
@@ -110,6 +112,10 @@ resource "google_compute_firewall" "allow_windows_activation" {
     protocol = "tcp"
     ports    = ["1688"]
   }
+
+  destination_ranges = ["35.190.247.13/32"]
+
+  target_tags = ["allow-win"]
```
rjerrems, line 118: I think something more specific like `allow-win-kms` or `allow-win-activation` would be better for this

3-networks/modules/standard_shared_vpc/firewall.tf
```diff
@@ -51,6 +51,8 @@ resource "google_compute_firewall" "allow_private_api_egress" {
   }
 
   destination_ranges = [local.private_googleapis_cidr]
+
+  target_tags = ["allow-google-api"]
```
rjerrems, line 55: nit: I think this should be `allow-google-apis`

3-networks/modules/standard_shared_vpc/firewall.tf
daniel-cit: I will user `allow-win-activation`

PR #181, opened 2020-08-04
fix multiple naming issues

daniel-cit:
- jenkins subnetwork
- logging project
- env code in common project
- dns hub project, vpc and subnets
- cloud routers
- interconnect vlans

### DIFF
--- a/1-org/envs/shared/projects.tf
+++ b/1-org/envs/shared/projects.tf
@@ -24,7 +24,7 @@ module "org_audit_logs" {
   random_project_id           = "true"
   impersonate_service_account = var.terraform_service_account
   default_service_account     = "depriviledge"
-  name                        = "prj-p-org-audit-logs"
+  name                        = "prj-c-logging"
   org_id                      = var.org_id
   billing_account             = var.billing_account
   folder_id                   = google_folder.common.id
@@ -33,7 +33,7 @@ module "org_audit_logs" {
 
   labels = {
     environment      = "prod"
-    application_name = "org-audit-logs"
+    application_name = "org-logging"
   }
 }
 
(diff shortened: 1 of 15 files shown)


1-org/envs/shared/projects.tf
```diff
@@ -43,7 +43,7 @@ module "org_billing_logs" {
   random_project_id           = "true"
   impersonate_service_account = var.terraform_service_account
   default_service_account     = "depriviledge"
-  name                        = "prj-p-org-billing-logs"
```
rjerrems, line 46: we should remove `-org-` from all of the projects, if we are doing it for the others

3-networks/modules/dedicated_interconnect/main.tf
```diff
@@ -26,15 +30,15 @@ module "interconnect_attachment1_region1" {
   source  = "terraform-google-modules/cloud-router/google//modules/interconnect_attachment"
   version = "~> 0.2.0"
 
-  name    = "vl-${var.region1_interconnect1_location}-${var.vpc_name}-${var.region1}-cr1"
+  name    = "vl-${var.region1_interconnect1_location}-${var.vpc_name}-${var.region1}-${local.suffix1}"
```
rjerrems, line 33: I think it would be better to explicitly provide the value, with a variable like `cloud_router_label`

daniel-cit: good idea

3-networks/modules/dedicated_interconnect/variables.tf
```diff
@@ -103,3 +103,9 @@ variable "region2_router2_name" {
   type        = string
   description = "Name of the Router 2 for Region 2 where the attachment resides"
 }
+
+variable "cloud_router_labels" {
+  type        = map(string)
+  description = "A map of suffixes for labelling vlans"
```
bharathkkb, line 109: nit: can we add what the structure should be? for example`"vlan_1" =>"suffix1"`, so the use knows what are the right keys. If not in the description, you can also add a comment in the code as well, just so its easier for the end user.

daniel-cit: Done

PR #198, opened 2020-08-05
fix: branch so that regex still works for cloud build with new environment names

rjerrems:
Currently the branch `non-production` triggers both apply and plan for cloud build, this change fixes that behavior. More details on bug raised on bootstrap module https://github.com/terraform-google-modules/terraform-google-bootstrap/issues/52

### DIFF
--- a/0-bootstrap/main.tf
+++ b/0-bootstrap/main.tf
@@ -162,7 +162,7 @@ module "cloudbuild_bootstrap" {
 
   terraform_apply_branches = [
     "development",
-    "non-production",
+    "non\\-production", //non-production needs a \ to ensure regex matches correct branches.
     "production"
   ]
 }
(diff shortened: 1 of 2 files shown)


0-bootstrap/main.tf
```diff
@@ -162,7 +162,7 @@ module "cloudbuild_bootstrap" {
 
   terraform_apply_branches = [
     "development",
-    "non-production",
+    "non\\-production", //non-production needs a \ to ensure regex matches correct branches.
```
caleonardo, line 165: In the whitepaper we used "nonprod" without a dash - Can we use "nonproduction" here to simplify the script?

rjerrems: the whitepaper lists the folder names as `non-production`, or are the branches listed differently?

rjerrems: In the case of Jenkins, I don't think this will change the behavior, this is to make the regex used in cloud build happy

PR #225, opened 2020-08-12
[Do Not Merge] Change default region for Jenkins agents to us-east4

morgante:
We would like to switch the default region to us-east4.

### DIFF
--- a/0-bootstrap/modules/jenkins-agent/variables.tf
+++ b/0-bootstrap/modules/jenkins-agent/variables.tf
@@ -36,7 +36,7 @@ variable "group_org_admins" {
 variable "default_region" {
   description = "Default region to create resources where applicable."
   type        = string
-  default     = "us-central1"
+  default     = "us-east4"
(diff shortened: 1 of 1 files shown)


0-bootstrap/modules/jenkins-agent/variables.tf
```diff
@@ -36,7 +36,7 @@ variable "group_org_admins" {
 variable "default_region" {
   description = "Default region to create resources where applicable."
   type        = string
-  default     = "us-central1"
+  default     = "us-east4"
```
bharathkkb, line 39: We can also use us-east1.

morgante:
Hi! It looks like you're updating the GCP Resource Hierarchy. Once this change is applied, your new hierarchy will look like this:

Screen Shot 2020-08-12 at 9 18 06 PM

rjerrems:
Does this PR need to stay open @morgante ?

PR #232, opened 2020-08-17
Feature/add integration tests 4 projects

amandakarina:
Hey folks

This pull request adds integration test for step 4-projects.


test/integration/projects/controls/development.rb
```diff
@@ -0,0 +1,137 @@
...
+    it { should exist }
+    its('lifecycle_state') { should cmp 'ACTIVE' }
+  end
+
+  describe google_project(project: dev_bu2_project_floating) do
+    it { should exist }
+    its('lifecycle_state') { should cmp 'ACTIVE' }
+  end
+
+  describe google_project(project: dev_bu2_project_restricted_id) do
```
bharathkkb, line 53: @amandakarina I think a good test would be if the projects are being attached as service projects to the correct host projects? 
Maybe using https://cloud.google.com/sdk/gcloud/reference/compute/shared-vpc/get-host-project or https://cloud.google.com/sdk/gcloud/reference/compute/shared-vpc/list-associated-resources to assert our logic of attachment and labels are right. WDYT?

amandakarina: Makes totally sense @bharathkkb! I'll add it!

rjerrems (approved):
LGTM - I think in future, we should look at consolidating / removing tests which are already run in the child modules (e.g project creation, API enablement) so that we can focus on testing things more specific to the foundation, but this is a more general comment about the integration tests.

PR #233, opened 2020-08-19
explicitely add project to scc pub/sub topic creation

camilojimenez:
I was having issues where `gcloud` was using the wrong project_id.


rjerrems (approved):
LGTM - Thank you @camilojimenez

Issue #227, opened 2020-08-14
0-Bootstrap with Jenkins - Automate the VPN creation with Terraform


caleonardo:
When implementing the 0-Bootstrap step using Jenkins, it is necessary to configure a VPN connection between the `prj-cicd` project (where the Jenkins Agent lives) and <----> on-prem (or where ever the Jenkins Master lives). Here are the [instructions to configure the VPN connection](https://github.com/terraform-google-modules/terraform-example-foundation/blob/master/0-bootstrap/README-Jenkins.md#iii-create-vpn-connection), which basically require to run several `gcloud` commands (which is easier than providing UI instructions).

This issue is to track the implementation of terraform scripts to automate the VPN creation. 
- Most likely, this VPN configuration will live in the [jenkins-agent/main.tf](https://github.com/terraform-google-modules/terraform-example-foundation/blob/master/0-bootstrap/modules/jenkins-agent/main.tf) file.
- An example configuration is in [3-networks/modules/vpn-ha](https://github.com/terraform-google-modules/terraform-example-foundation/tree/master/3-networks/modules/vpn-ha)

atos-cit:
Done in PR #234

PR #234, opened 2020-08-25
Adds jenkins agent vpn terraform scripts

atos-cit:
Related with issue #227. Adds HA VPN code, instructions in README and removes gcloud commands to create Classic VPN.


rjerrems (commented):
Thanks @atos-cit looks like docs need to be regenerated.

Do we really need to have examples for both standard and HA VPN? Could we not just link to the example for the module if the customer wants HA VPN? The reason I say this, is this solution for on-prem access should only be temporary until the interconnect / partner interconnect is stood up and would be good to not maintain additional variations where possible.

atos-cit:
> Thanks @atos-cit looks like docs need to be regenerated.
> 
> Do we really need to have examples for both standard and HA VPN? Could we not just link to the example for the module if the customer wants HA VPN? The reason I say this, is this solution for on-prem access should only be temporary until the interconnect / partner interconnect is stood up and would be good to not maintain additional variations where possible.

Hi @rjerrems, thanks for the feedback. I made those changes based on an discussion that Carlos brought in Chats about what type of VPN we wanted to add. Could we add a note in the readme citing that we already provide an example code and directing to official documentation?

rjerrems:
Thanks for the context @atos-cit - I am still leaning towards only maintaining one version. @bharathkkb @morgante what do you think?

morgante:
Yeah I'm in favor of 1 version as well (probably HA).

atos-cit:
@rjerrems @morgante I removed the Classic VPN. Please take a look if the solution is ok now.

0-bootstrap/README-Jenkins.md
```diff
@@ -157,101 +156,16 @@ You arrived to these instructions because you are using the `jenkins_bootstrap`
...
-
-    # Create the necessary Route
-    gcloud compute --project $CICD_PROJECT_ID \
-      routes create "${CICD_VPN_NAME}-tunnel-1-route-1" \
-      --network $JENKINS_AGENT_VPC_NAME \
-      --next-hop-vpn-tunnel "${CICD_VPN_NAME}-tunnel-1" \
-      --next-hop-vpn-tunnel-region $DEFAULT_REGION \
-      --destination-range $JENKINS_MASTER_NETWORK_CIDR
-    ```
+1. Check in the `prj-cicd` project which are the gateway static IP address that has been reserved for your VPN. You **NEED** to inform your Network administrator those IP address so they configure the on-prem side of the VPN tunnels.
```
rjerrems, line 168:
```suggestion
1. Check in the `prj-cicd` project for the VPN gateway static IP addresses which have been reserved. These addresses are required by the Network Administrator for the configuration of the on-prem side of the VPN tunnels to GCP.
```

rjerrems:
Overall looking good, just one minor nit on the docs @atos-cit - then should be good to merge from my point of view.

rjerrems (approved):
LGTM

cc @bharathkkb

PR #239, opened 2020-09-03
Adds explanation for optional variables

atos-cit:
Adds explanation for optional variables in steps 0, 1 and 2

### DIFF
--- a/0-bootstrap/README-Jenkins.md
+++ b/0-bootstrap/README-Jenkins.md
@@ -286,9 +286,7 @@ Here you will configure a VPN Network tunnel to enable connectivity between the 
 
 ### Optional Variables
 
-Some variables used to deploy the step have default values. Check if you want to change those values before doing the deploy:
-
-- **skip_gcloud_download:** By default this option is `true`. If you change to false it will download gcloud. The default value assumes gcloud is already available outside the module.
+Some variables used to deploy the step have default values, check if you want to change those values before doing the deploy. For more information check in the section [Inputs](./modules/jenkins-agent/README.md#Inputs) the variables not required.
 
 ## Contributing
 
--- a/0-bootstrap/README-Jenkins.md
+++ b/0-bootstrap/README-Jenkins.md
@@ -113,7 +113,6 @@ You arrived to these instructions because you are using the `jenkins_bootstrap` 
     1. Un-comment the `jenkins_bootstrap` module in `./main.tf`
     1. Un-comment the `jenkins_bootstrap` variables in `./variables.tf`
     1. Un-comment the `jenkins_bootstrap` outputs in `./outputs.tf`
-
 1. Rename `terraform.example.tfvars` to `terraform.tfvars` and update the file with values from your environment.
     - One of the value to supply (variable `jenkins_agent_gce_ssh_pub_key`) is the **public SSH key** you generated in the first step.
         - **Note: this is not the secret private key**. The public SSH key can be in your repository code.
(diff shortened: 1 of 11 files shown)


0-bootstrap/README-Jenkins.md
```diff
@@ -284,6 +284,12 @@ Here you will configure a VPN Network tunnel to enable connectivity between the
 
 1. You can now move to the instructions in the step [1-org](../1-org/README.md).
 
+### Optional Variables
+
+Some variables used to deploy the step have default values. Check if you want to change those values before doing the deploy:
+
+- **skip_gcloud_download:** By default this option is `true`. If you change to false it will download gcloud. The default value assumes gcloud is already available outside the module.
```
caleonardo, line 291: It is usually easier to follow positive statements like:
`download_gcloud = true` -> then it means `gcloud` will be downloaded.

So, is `skip_gcloud_download`is only used here and it is not a variable that we use in other modules, I would change `skip_gcloud_download` to `download_gcloud`

atos-cit: I agree, but I think would be better to fix this in another PR. Thanks!

bharathkkb: `skip_gcloud_download` is from `gcloud` module which exposes `skip_download`. I think we kept it because default behaviour of gcloud module was to enable users to download and use gcloud. With the 2.0 release we default `skip_download` to true and expose more controls, so we wont need to plumb this. But yes, I agree with @atos-cit we can take care of that bump and minor refactor in a separate PR.

0-bootstrap/README-Jenkins.md
```diff
@@ -284,6 +284,12 @@ Here you will configure a VPN Network tunnel to enable connectivity between the
 
 1. You can now move to the instructions in the step [1-org](../1-org/README.md).
 
+### Optional Variables
```
bharathkkb, line 287: This readme is getting pretty big. Thoughts on a Table of contents with anchors?

1-org/README.md
```diff
@@ -72,3 +72,30 @@ When using Cloud Build or Jenkins as your CI/CD tool each environment correspond
 1. Run `./tf-wrapper.sh apply production`
 
 If you received any errors or made any changes to the Terraform config or `terraform.tfvars` you must re-run `./tf-wrapper.sh plan production` before run `./tf-wrapper.sh apply production`
+
+### Optional Variables
+
+Some variables used to deploy the step have default values. Check if you want to change those values before doing the deploy:
+
+- **skip_gcloud_download:** By default this option is `true`. The default value assumes gcloud is already available outside the module, if you change to false it will download gcloud.
```
bharathkkb, line 80: From a maintainability perspective, I dont think handwriting these are really sustainable. I feel like we can either
- highlight important ones
- use a custom [terraform-docs](https://github.com/terraform-docs/terraform-docs/blob/master/docs/CONFIG_FILE.md) config to autogen some version of this.

bharathkkb (commented):
@rjerrems thoughts on these?

1-org/README.md
rjerrems: +1 we should definitely auto-generate these docs with the same convention we are using elsewhere. There should be a better way to highlight optional variables

0-bootstrap/README-Jenkins.md
rjerrems: +1

rjerrems:
Agree @bharathkkb - we should definitely be generating docs, otherwise it will easily get out of sync with the code. There are probably a few ways we can make it clearer which variables are optional, potentially in the description.

atos-cit:
@bharathkkb @rjerrems, I took a look at how the scripts generated the docs in the docker image  (also @bharathkkb provided us later with the link to the repo). I observed that:
- We would need to make a different script since the one used by the docker image uses the `terraform-docs` to create a table of inputs and outputs for each terraform module, and we are highlighting all the optional variables in the top-level README.
- It would be non-trivial to change since it needs to get all the descriptions from different modules, and the `terraform-docs` generates the inputs and outputs table for a module.

Since the final result using terraform-docs would be a table with the contents of the input variables just like the ones on each generated in each terraform module, I suggest that we link to those respective READMEs that already exists and have descriptions of those variables. This would avoid the duplication of those explanations and the need for a new script to generate documentation.

bharathkkb:
@atos-cit that approach SGTM. I think the purpose is to just call to attention that we are indeed setting these and that the user can change them if necessary.

atos-cit:
@rjerrems @bharathkkb Any new thoughts on the proposed solution?

0-bootstrap/README-Jenkins.md
```diff
@@ -198,6 +198,10 @@ Here you will configure a VPN Network tunnel to enable connectivity between the
...
+
+Some variables used to deploy the step have default values, check if you want to change those values before doing the deploy. For more information check in the section [Inputs](./modules/jenkins-agent/README.md#Inputs) the variables not required.
```
rjerrems, line 203:
```suggestion
Some variables used to deploy this step have default values, please check these prior to deployment to ensure they match your requirements. For more information, see [Inputs](./modules/jenkins-agent/README.md#Inputs) for variables marked as not required.
```

0-bootstrap/README.md
```diff
@@ -37,6 +37,10 @@ Currently, the bucket information is replaced in the state backends as a part of
 1. Change into the main directory for the terraform-example-foundation.
 1. Run this command ```for i in `find -name 'backend.tf'`; do sed -i 's/UPDATE_ME/GCS_BUCKET_NAME/' $i; done``` where `GCS_BUCKET_NAME` is the name of your bucket from the steps executed above.
 
+### Optional Variables
+
+Some variables used to deploy the step have default values, check if you want to change those values before doing the deploy. For more information check in the section [Inputs](./README.md#Inputs) the variables not required.
```
rjerrems, line 42:
```suggestion
Some variables used to deploy this step have default values, please check these prior to deployment to ensure they match your requirements. For more information, see [Inputs](./README.md#Inputs) for variables marked as not required.
```

rjerrems:
Hi @atos-cit - I agree with the approach, but would it be better to add a statement in one place rather than adding it in many different places that we then need to keep up to date? I am thinking something generic in the main README, similar to branching strategy might make sense? https://github.com/terraform-google-modules/terraform-example-foundation#branching-strategy

atos-cit:
Thanks for the feedback @rjerrems ! I agree with you, it would be easier to maintain. I will made the necessary changes.

rjerrems (approved):
LGTM

cc @bharathkkb

PR #248, opened 2020-09-28
[WIP] Centralize configuration in a constants module.

### LINKED ISSUE #244: [devex] Improve data gathering per stage
At each stage, we ask the user to populate `terraform.tfvars` with values from previous stages or static data like `billing_account`. This seems like repeated data gathering and we should provide an alternative to automate/script some of this.

morgante:
This introduces a "constants" module which can be used to centralize most configuration instead of repeating variables throughout the codebase.

Fixes #244.

### DIFF
--- a/0-bootstrap/README.md
+++ b/0-bootstrap/README.md
@@ -21,7 +21,6 @@ If you are using the `jenkins_bootstrap` sub-module, please see [README-Jenkins]
 ## 0-bootstrap usage to deploy Cloud Build
 
 1. Change into 0-bootstrap folder
-1. Copy tfvars by running `cp terraform.example.tfvars terraform.tfvars` and update `terraform.tfvars` with values from your environment.
 1. Run `terraform init`
 1. Run `terraform plan` and review output
 1. Run `terraform apply`
--- a/0-bootstrap/main.tf
+++ b/0-bootstrap/main.tf
@@ -63,7 +63,7 @@ module "seed_bootstrap" {
 
   default_region = module.constants.values.default_region
 
-  org_project_creators = module.constants.values.bootstrap.org_project_creators
+  org_project_creators = lookup(module.constants.values.bootstrap, "org_project_creators", [])
 
   sa_enable_impersonation = true
   skip_gcloud_download    = true
--- a/0-bootstrap/main.tf
+++ b/0-bootstrap/main.tf
@@ -31,10 +31,17 @@ provider "random" {
 }
 
 /*************************************************
+  Import Constants.
+*************************************************/
+module "constants" {
+  source = "../modules/constants"
+}
+
+/*************************************************
   Bootstrap GCP Organization.
 *************************************************/
 locals {
-  parent = var.parent_folder != "" ? "folders/${var.parent_folder}" : "organizations/${var.org_id}"
+  parent = module.constants.deploy_at_root ? "organizations/${module.constants.values.org_id}" : "folders/${module.constants.values.parent_folder}"
 }
 
 resource "google_folder" "bootstrap" {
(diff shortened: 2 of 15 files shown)


morgante:
@bharathkkb @rjerrems What do you think of this approach?

rjerrems:
I like the idea @morgante - my main reservation is that is won't be particularly valuable unless you have a mono repo (the default at the moment is multi-repo). Unless I am missing how you would prevent having to copy it in a multi-repo config?

bharathkkb:
I like the idea as well.

> my main reservation is that is won't be particularly valuable unless you have a mono repo

@rjerrems perhaps in the  multi-repo approach we can deploy this to a `shared-modules` repo and source via git?
Another approach could also be to store this in `0-bootstrap/modules` and reference else where as `source = git::foo/bootstrap.git//modules/constants`.

morgante:
Yeah, for multiple repos there are a few options:

1. Store it in a shared module repo (or one of the repos) and reference via Git.
2. Expose the module outputs as outputs from 0-bootstrap and use remote state to reference them from the other repos.

rjerrems:
Yeah cool, I am happy so long as we continue to support multi-repo through one of those mechanisms. (whatever is lower friction for users/customers)

morgante:
I think the remote state mechanism is likely lower friction.

rjerrems:
There should only actually be one state bucket and if we push the state in the bootstrap stage, it should be re-usable at the same path / bucket name throughout. That approach sounds viable to me

rjerrems:
Hows this going? Did you need some help to finish this off @morgante ?

morgante:
Yeah I've been busy with other things / mostly just meant to demonstrate the approach. If someone could pick this up I'd definitely appreciate the help.

umairidris:
I have been exploring this idea as well in some of our solutions and definitely think it is a good idea. Some modifications I made that could be considered here:

- The constants are returned as a map of env to values

```
output "values" {
  value = {
    dev = {
      env_code  = "d"
      folder_id = "456"
    }
    shared = {
      billing_account = "000"
      env_code        = "s"
      folder_id       = "123"
      project_prefix  = "example-prefix"
      region          = "us-central1"
      state_bucket    = "example-state"
    }
  }
}
```

- Every env passes in a name which corresponds to a name in the constants map (e.g. 'dev')

```
module "main" {
  source = "../../modules/main"
  env    = "dev"
}
```

- Modules will then import the constants and access values as needed

```
module "constants" {
  source = "../../../constants"
}

locals {
  constants = merge(module.constants.values.shared, module.constants.values[var.env])
  project_id = "${local.constants.project_prefix}-fooproj-${local.constants.env_code}"
}

module "bucket" {
  source  = "terraform-google-modules/cloud-storage/google//modules/simple_bucket"
  version = "~> 1.4"

  name        = "${local.constants.project_prefix}-foobucket-${local.constants.env_code}"
  project_id = local.project_id
  location    = local.constants.region
}
```

gaspar-chilingarov:
My 2 cents :))) 

I would go with remote state approach, because it allows keep stages 1,2,3 exactly the same, even if the source code is shared between different GCP organisations (sandbox/production) - all environment specific configuration is pulled form bootstrap. (I would put there project prefix name too). 

About the bucket name - I personally solved this with removing bucket reference :) 

```
terraform {
  backend "gcs" {
    prefix = "org-name-stage02"
  }
}
```

and running terraform init as 
```
terraform init -backend-config=bucket=${TF_STATE_BUCKET}
```

Same applies to remote state
```
data "terraform_remote_state" "stage01" {
  backend = "gcs"

  config = {
    prefix = "org-name-stage01"
    bucket = var.host_state_bucket
  }
}
```
And supplying the variable from the command line 

```
terraform apply -var="host_state_bucket=${TF_STATE_BUCKET}"
```

In that way TF files are completely unaware in which environment they are running and correct state bucket can be fed from the ENV/pipeline configuration. In that way between multiple orgs TF files remain exactly same, only the CB configuration changes.

What do you think?

PR #252, opened 2020-09-30
[WIP] Set hipaa audit controls

umairidris:
HIPAA requires 6 years of audit log retention (see e.g. https://www.aptible.com/documentation/faq/comply/hipaa-audit-logs.html). Thus, we set the retention to 7 years forGCS to ensure compliance and use COLDLINE storage class for cost efficiency. 

We use BigQuery for general log querying and thus set this to 1 year retention.

Dependencies:
https://github.com/terraform-google-modules/terraform-google-log-export/pull/63
https://github.com/terraform-google-modules/terraform-google-log-export/pull/64

### DIFF
--- a/1-org/envs/shared/log_sinks.tf
+++ b/1-org/envs/shared/log_sinks.tf
@@ -39,7 +39,7 @@ resource "random_string" "suffix" {
 
 module "log_export_to_biqquery" {
   source                 = "terraform-google-modules/log-export/google"
-  version                = "~> 4.0"
+  version                = "~> 5.0"
   destination_uri        = module.bigquery_destination.destination_uri
   filter                 = local.all_logs_filter
   log_sink_name          = "sk-c-logging-bq"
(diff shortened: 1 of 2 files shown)


morgante:
We should probably start a HIPAA branch for now, as I'm not sure that we want to incorporate all HIPAA requirements into the baseline yet.

umairidris:
> We should probably start a HIPAA branch for now, as I'm not sure that we want to incorporate all HIPAA requirements into the baseline yet.

Auditing is probably the main part of infra that has specific requirements for HIPAA. Other regulations are more generic such as principle of least privilege, minimizing human access, reducing public internet access, encryption, etc which are mostly already set in the foundation and just require a few tweaks.

morgante:
> Auditing is probably the main part of infra that has specific requirements for HIPAA. Other regulations are more generic such as principle of least privilege, minimizing human access, reducing public internet access, encryption, etc which are mostly already set in the foundation and just require a few tweaks.

Yeah, I imagine the branch would be mostly the same. The main reason I think we should consider a branch is that this PR probably shouldn't be merged into `master` currently.

> I can also add a doc to the foundation for HIPAA controls if you'd like? In the future we can expand this to other compliance regulations that we are aligning with.

I think a convergence doc is definitely a good idea, but we don't need to add much more to the repo until we're agreed on the path forward.

rjerrems:
Closing this for now as it looks stale, feel free to re-open if this is active.

PR #261, opened 2020-10-05
Removes filter for sinks

### LINKED ISSUE #258: [shascanner] LOG_NOT_EXPORTED detected
LOG_NOT_EXPORTED is detected because log sinks are created with 
    logName: /logs/cloudaudit.googleapis.com%2Faccess_transparency

And the detector considers that the sink should have an empty filter to log every possible log created.

Changes sink creation to use a empty filter.

amandakarina:
Hey folks

This pull request closes #258 issue.

Thanks!

### DIFF
--- a/1-org/README.md
+++ b/1-org/README.md
@@ -15,6 +15,10 @@ Enabling Data Access logs might result in your project being charged for the add
 For details on costs you might incur, go to [Pricing](https://cloud.google.com/stackdriver/pricing).
 You can choose not to enable the Data Access logs by setting variable `data_access_logs_enabled` to false.
 
+**Note:** This module creates sink to export all logs to Google Storage. It's also create sinks to export the main
+logs to Bigquery and Pub/Sub. This might result in your project being charged.
+You can change those filters if you want on the file `envs/shared/log_sinks.tf`.
+
 **Note:** Currently, this module does not enable bucket policy retention for organization logs, please, enable it if needed.
 
 ### Setup to run via Cloud Build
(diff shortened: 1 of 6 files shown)


rjerrems:
Hi @amandakarina - would be good to understand some extra context on what we are trying to achieve here, I have commented on the linked bug

amandakarina:
> Hi @amandakarina - would be good to understand some extra context on what we are trying to achieve here, I have commented on the linked bug

Hey @rjerrems 
The SHA scanner detected this finding after deploy CFT in a clean organization. The SHA Scanner checks if the sink has a empty filter, if so, no finding is generated, otherwise, a LOG_NOT_EXPORTED finding is created.

rjerrems:
Yep - I suppose the question I have is: are we recommending that people export *all* logs? Are we confident the SHA finding is representative of best practice here? It doesn't seem to stack up with my experience where specific log types are usually targeted. It will also increase cost, particularly in our case where we are creating three copies of the exported logs to GCS, BigQuery & PubSub.

amandakarina:
Hey @rjerrems, I remove the filter only for bucket export. Could you, please take a look?

rjerrems:
Thanks @amandakarina - can we add some brief docs that explain the difference, then we should be good to go.

Just to confirm also, that the SHA scanner result is fixed by this change?

amandakarina:
> Thanks @amandakarina - can we add some brief docs that explain the difference, then we should be good to go.
> 
> Just to confirm also, that the SHA scanner result is fixed by this change?

Hey Rohan!

Based on my test, yes!
I inactivated all findings for LOG_NOT_EXPORTED and create a new project under my folder.
After SHA Scanner was executed no new finding was created and no finding for my folder was re-activated.

rjerrems (approved):
LGTM

rjerrems:
FYI Merging because test failure is due to name collision & healthy test run on prior commit

PR #262, opened 2020-10-07
Adds org policy admin role for admin group

### LINKED ISSUE #254: [shascanner] Add org policy admin role to Admins
Add org policy admin role to Admins

amandakarina:
Hey folks!

This pull request closes #254 issue.

Thanks!

### DIFF
--- a/0-bootstrap/README.md
+++ b/0-bootstrap/README.md
@@ -47,7 +47,7 @@ Currently, the bucket information is replaced in the state backends as a part of
 | group\_billing\_admins | Google Group for GCP Billing Administrators | string | n/a | yes |
 | group\_org\_admins | Google Group for GCP Organization Administrators | string | n/a | yes |
 | org\_id | GCP Organization ID | string | n/a | yes |
-| org\_policy\_admin\_role | Additional Org Policy Admin role for admin group. You can use this for testing prourpose. | bool | `"false"` | no |
+| org\_policy\_admin\_role | Additional Org Policy Admin role for admin group. You can use this for testing purposes. | bool | `"false"` | no |
 | org\_project\_creators | Additional list of members to have project creator role across the organization. Prefix of group: user: or serviceAccount: is required. | list(string) | `<list>` | no |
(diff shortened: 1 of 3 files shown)


bharathkkb (approved):
LGTM
Just a question @rjerrems is this something we should upstream to bootstrap as one of the defaults?

rjerrems:
Good question @bharathkkb - part of the reason it is not in there is hesitation on part. Preferably IMHO, you would avoid cases where people can modify organization policy outside of code and keep the permissions assigned to the minimum required to bootstrap the org. (i.e the tf service account applies org policy thereafter)

What are the main use cases we are trying to fulfill by adding the permissions for the group here? Is it mainly for convenience?

morgante:
+1 to @rjerrems, I think by default we should leave this off.

bharathkkb:
This makes sense.  I will circle back with the team to get specifics on why the SHA scanner complained.

0-bootstrap/README.md
```diff
@@ -47,6 +47,7 @@ Currently, the bucket information is replaced in the state backends as a part of
 | group\_billing\_admins | Google Group for GCP Billing Administrators | string | n/a | yes |
 | group\_org\_admins | Google Group for GCP Organization Administrators | string | n/a | yes |
 | org\_id | GCP Organization ID | string | n/a | yes |
+| org\_policy\_admin\_role | Additional Org Policy Admin role for admin group. You can use this for testing prourpose. | bool | `"false"` | no |
```
rjerrems, line 50: nit: typo for `purpose`

rjerrems (approved):
LGTM

rjerrems:
FYI @morgante @bharathkkb - the behavior now is to allow people to enable the policy admin role via a flag, which defaults to false. (not adding those permissions)

Call out if you have any reservations otherwise will merge.

bharathkkb:
@rjerrems sounds good, the flagged approach is what we agreed on offline.

PR #271, opened 2020-10-23
added Shielded VMs constraint

daniel-cit:
Add [Shielded VMs](https://cloud.google.com/security/shielded-cloud/shielded-vm) constraint to the list of org policies

### DIFF
--- a/1-org/envs/shared/org_policy.tf
+++ b/1-org/envs/shared/org_policy.tf
@@ -91,6 +91,17 @@ module "org_shared_vpc_lien_removal" {
   constraint      = "constraints/compute.restrictXpnProjectLienRemoval"
 }
 
+module "org_shared_require_shielded_vm" {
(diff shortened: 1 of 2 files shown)


bharathkkb (approved):
LGTM

rjerrems:
Do we know what the impact of applying this policy is for service that launch VMs in the customer project? e.g Dataflow, Composer etc - do these services support shielded VMs?

bharathkkb:
@rjerrems good catch, both don't seem supported. Perhaps we should move this to a note like are doing for `oslogin`.

rjerrems:
Yeah I think that might be worthwhile, otherwise customers will be knee deep in project exemptions to use a number of services.(although some customers may want this)

daniel-cit:
New PR create for develop branch with TF 0.13

PR #277, opened 2020-10-30
feat: Make BigQuery log destinations to be partitioned

### LINKED ISSUE #273: Convert BQ log destinations are configured to be partitioned
In the current code, we configure BQ log exports as sharded tables; we should modify this to create them as partitioned tables, which are more user-friendly and performant.

This fix can be implemented by adding the following to the `log_export_to_bigquery` [module](https://github.com/terraform-google-modules/terraform-example-foundation/blob/5c5b8b377494751cae70314976420ab04ddc5e55/1-org/envs/shared/log_sinks.tf#L41)

```
  bigquery_options       = {
    use_partitioned_tables  = true
  }
```

vovinacci:
Closes #273

### DIFF
--- a/1-org/envs/shared/log_sinks.tf
+++ b/1-org/envs/shared/log_sinks.tf
@@ -35,12 +35,15 @@ resource "random_string" "suffix" {
 }
 
 /******************************************
-  Send logs to BigQury
+  Send logs to BigQuery
 *****************************************/
 
 module "log_export_to_biqquery" {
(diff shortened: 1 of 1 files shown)


vovinacci:
Integration tests are failing, but I don't have access to check what is failing exactly.

bharathkkb (approved):
Thanks for the PR @vovinacci!
/cc @rjerrems @mikelaramie

PR #283, opened 2020-11-05
New org policies

daniel-cit:
This PR adds two new org policies:

- OS login 
- Shielded VMs

with a toggle for each one

### DIFF
--- a/1-org/envs/shared/variables.tf
+++ b/1-org/envs/shared/variables.tf
@@ -49,6 +49,18 @@ variable "domains_to_allow" {
   type        = list(string)
 }
 
+variable "enable_shielded_vm_policy" {
+  description = "Enable Shielded VMs policy."
+  type        = bool
+  default     = false
+}
+
+variable "enable_os_login_policy" {
+  description = "Enable OS Login policy."
+  type        = bool
+  default     = false
+}
+
 variable "audit_logs_table_expiration_days" {
   description = "Period before tables expire for all audit logs in milliseconds. Default is 30 days."
   type        = number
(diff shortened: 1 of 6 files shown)


1-org/envs/shared/variables.tf
```diff
@@ -49,6 +49,18 @@ variable "domains_to_allow" {
   type        = list(string)
 }
 
+variable "enable_shielded_vm_policy" {
```
bharathkkb, line 52: @daniel-cit can we document the caveats for enabling each of these as well (i.e GKE issue with os login and dataflow/composer with Shielded VMs) so that the user is aware.

bharathkkb (commented):
overall LGTM
/cc @rjerrems

rjerrems:
So the plan is turn these on by default now? We should have some documentation for how to add exceptions to the policy for workloads that don't work well with these policies I think before we make it the default

bharathkkb:
@rjerrems off by default and +1 for docs.
https://github.com/terraform-google-modules/terraform-example-foundation/blob/3ce143ade1ef66bfef004375777736c2b3509444/1-org/envs/shared/variables.tf#L55

rjerrems (approved):
LGTM

1-org/envs/shared/variables.tf
rjerrems: I think its best to link to public docs as @daniel-cit has done, so that they can be the source of truth for any limitations

rjerrems: (we should try to get docs updated where they are lacking)

PR #289, opened 2020-11-16
feat: Adds prefix to projects and folder name

### LINKED ISSUE #270: Add global prefix for created projects
As users are now deploying example-foundation, they will run into issues of reused project id's just like we do in CI. I propose we add a global prefix option, default to `""` and override default to `"test-"` for all of our CI tests. This will ensure less flakiness from reused project ids as well as give customers ability to add a prefix if needed.

### LINKED ISSUE #275: Move folder prefix to variable
Similarly to #270 , if an organization chooses to change or remove the `fldr-` prefix from their folder names, it entails modifying a good bit of code.  This prefix (and others) should be moved to a variable to make it easier for an organization to change to their own potentially pre-defined standards.

amandakarina:
Hey folks,

This pull request closes #270 issue and closes #275 issue.

Thanks!

### DIFF
--- a/0-bootstrap/variables.tf
+++ b/0-bootstrap/variables.tf
@@ -64,6 +64,12 @@ variable "skip_gcloud_download" {
   default     = true
 }
 
+variable "project_prefix" {
+  description = "Name prefix to use for projects created."
+  type        = string
+  default     = "prj"
+}
+
 /* ----------------------------------------
     Specific to jenkins_bootstrap module
    ---------------------------------------- */
--- a/0-bootstrap/variables.tf
+++ b/0-bootstrap/variables.tf
@@ -70,6 +70,12 @@ variable "project_prefix" {
   default     = "prj"
 }
 
+variable "folder_prefix" {
+  description = "Name prefix to use for folders created."
+  type        = string
+  default     = "fldr"
+}
+
 /* ----------------------------------------
     Specific to jenkins_bootstrap module
    ---------------------------------------- */
--- a/2-environments/envs/development/variables.tf
+++ b/2-environments/envs/development/variables.tf
@@ -39,3 +39,9 @@ variable "monitoring_workspace_users" {
   description = "Gsuite or Cloud Identity group that have access to Monitoring Workspaces."
   type        = string
 }
+
+variable "project_prefix" {
+  description = "Name prefix to use for projects created."
+  type        = string
+  default     = "prj"
+}
--- a/2-environments/envs/development/variables.tf
+++ b/2-environments/envs/development/variables.tf
@@ -45,3 +45,9 @@ variable "project_prefix" {
   type        = string
   default     = "prj"
 }
+
+variable "folder_prefix" {
+  description = "Name prefix to use for folders created."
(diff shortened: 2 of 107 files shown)


test/fixtures/bootstrap/main.tf
```diff
@@ -22,4 +22,5 @@ module "bootstrap" {
   group_billing_admins = var.group_email
   billing_account      = var.billing_account
   org_project_creators = var.org_project_creators
+  project_prefix       = "tst"
```
daniel-cit, line 25: I think that to help maintainability, instead of using

```
project_prefix       = "tst"
```

here and in the other fixture files we cloud use 

```
project_prefix       = var.project_prefix
```

and set  `tst` only in the cloud build file

https://github.com/terraform-google-modules/terraform-example-foundation/blob/master/build/int.cloudbuild.yaml

what do you think @bharathkkb

daniel-cit:
When creating a project, setting the variable to empty:

```
project_prefix       = ""
```

will result in the error: **Project ID can have lowercase letters, digits or hyphens and must start with a lowercase letter.**


should we add a warning  in the variable description saying that the value **must** not be empty ?

**OR** do we need to write some logic around it?

@amandakarina @bharathkkb

0-bootstrap/variables.tf
```diff
@@ -64,6 +64,18 @@ variable "skip_gcloud_download" {
   default     = true
 }
 
+variable "project_prefix" {
```
bharathkkb, line 67: We should override these via the cloudbuild yaml for our integration tests

bharathkkb (changes requested):

0-bootstrap/variables.tf
bharathkkb: @daniel-cit already mentioned it here, so tracking https://github.com/terraform-google-modules/terraform-example-foundation/pull/289#discussion_r525124109

2-environments/envs/development/variables.tf
```diff
@@ -39,3 +39,15 @@ variable "monitoring_workspace_users" {
   description = "Gsuite or Cloud Identity group that have access to Monitoring Workspaces."
   type        = string
 }
+
+variable "project_prefix" {
```
bharathkkb, line 43: @amandakarina how are we plumbing this through to env_baseline? Shouldn't we pass these in here?
https://github.com/amandakarina/terraform-example-foundation/blob/feature/adds-project-prefix-variable/2-environments/envs/development/main.tf

amandakarina: Ops! It was missing! Thanks @bharathkkb

bharathkkb:
@amandakarina looking at the build output, it seems like in 4-projects we are still using `prj` for testing. Could PTAL

amandakarina:
> @amandakarina looking at the build output, it seems like in 4-projects we are still using `prj` for testing. Could PTAL

Thanks Bharath! It's fixed!

bharathkkb (approved):
LGTM
/cc @rjerrems

PR #296, opened 2020-12-02
Add projects for base/restricted network hubs.

averbuks:


test/integration/org/controls/gcp_projects.rb
```diff
@@ -82,6 +84,18 @@
     its('lifecycle_state') { should cmp 'ACTIVE' }
   end
 
+  describe google_project(project: base_net_hub_project_id) do
```
bharathkkb, line 87: These may fail if not in hub and spoke mode. I am thinking we can add a var for the hub and spoke fixture flag 
https://github.com/terraform-google-modules/terraform-example-foundation/blob/cf8fbd6766fbeb70c3267dda3364d2e1b563dc46/test/fixtures/org/main.tf#L38
and pass in an env var via cloudbuild to flip it like this
https://github.com/terraform-google-modules/terraform-example-foundation/blob/d6007d456ceb1b2ddf7e21a842ef5083bbe5586c/build/int.cloudbuild.yaml#L49

We can plumb this var back to fixture outputs then assert in the tests to execute these only if `hub_and_spoke` attribute is `true`.

I will put some thought into #297 this week but this should give us the flexibility. WDYT?

bharathkkb (approved):
Thanks @averbuks! LGTM
Tracking full int tests here https://github.com/terraform-google-modules/terraform-example-foundation/issues/297 so we can test both codepaths concurrently 

/cc @rjerrems

test/integration/org/controls/gcp_projects.rb
averbuks: Hi @bharathkkb, good one, I updated the tests and now hub_and_spoke can be true or false (true in our cloudbuild config). 
In case of true - we assert projects are created
In case of false - we assert output values for these projects are nulls, meaning projects are not created.

PR #300, opened 2020-12-08
Move source repository list to variables

jaroslawgajewski:


bharathkkb (changes requested):
Hi @jaroslawgajewski 
Thanks for the PR; looks like some unrelated changes are in this branch, could base off of `develop` and add your commits?

Issue #288, opened 2020-11-14
use module for_each for org policies


bharathkkb:
Current org policy config is verbose and could be refactored to be more concise using module for_each. This also allows end users to more easily see currently enforced policies and add new ones.

https://github.com/terraform-google-modules/terraform-example-foundation/blob/master/1-org/envs/shared/org_policy.tf

bharathkkb:
closing as per https://github.com/terraform-google-modules/terraform-example-foundation/pull/311#issuecomment-754285308

PR #311, opened 2020-12-21
Refactor org policies configuration

meganzhao10:
PR for #288: refactored org policies config to use for_each for more concise and readable code

### DIFF
--- a/1-org/README.md
+++ b/1-org/README.md
@@ -23,7 +23,7 @@ You can change the filters & sinks by modifying the configuration in `envs/share
 
 **Note:** It is possible to enable an organization policy for [OS Login](https://cloud.google.com/compute/docs/oslogin/manage-oslogin-in-an-org) with this module.
 OS Login has some [limitations](https://cloud.google.com/compute/docs/instances/managing-instance-access#limitations).
-If those limitations do not apply to your workload/environment you can choose to enable the OS Login policy by setting variable `enable_os_login_policy` to `true`.
+If those limitations do not apply to your workload/environment you can choose to enable the OS Login policy by uncommenting the "constraints/compute.requireOsLogin" policy constraint in `envs/shared/org_policy.tf` line 28.
 
 ### Setup to run via Cloud Build
 1. Clone repo `gcloud source repos clone gcp-org --project=YOUR_CLOUD_BUILD_PROJECT_ID` (this is from terraform output from the previous section, 0-bootstrap).
--- a/1-org/envs/shared/org_policy.tf
+++ b/1-org/envs/shared/org_policy.tf
@@ -25,7 +25,11 @@ locals {
     "constraints/compute.vmExternalIpAccess"            = "list",
     "constraints/compute.skipDefaultNetworkCreation"    = "boolean",
     "constraints/compute.restrictXpnProjectLienRemoval" = "boolean",
-    "constraints/compute.requireOsLogin"                = "boolean",
+    # "constraints/compute.requireOsLogin"                = "boolean",
+    # OS Login is disabled due to https://cloud.google.com/compute/docs/instances/managing-instance-access#limitations
+    "constraints/sql.restrictPublicIp"                 = "boolean",
+    "constraints/iam.disableServiceAccountKeyCreation" = "boolean",
+    "constraints/storage.uniformBucketLevelAccess"     = "boolean",
   }
 }
 
@@ -46,21 +50,6 @@ module "org_compute_policies" {
 }
 
 /******************************************
-  Cloud SQL
-*******************************************/
-
-module "org_cloudsql_external_ip_access" {
-  source          = "terraform-google-modules/org-policy/google"
-  version         = "~> 3.0"
-  organization_id = local.organization_id
-  folder_id       = local.folder_id
-  policy_for      = local.policy_for
-  policy_type     = "boolean"
-  enforce         = "true"
-  constraint      = "constraints/sql.restrictPublicIp"
-}
-
-/******************************************
   IAM
 *******************************************/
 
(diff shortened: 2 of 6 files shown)


1-org/envs/shared/org_policy.tf
```diff
@@ -18,89 +18,31 @@ locals {
...
-  folder_id       = local.folder_id
-  policy_for      = local.policy_for
-  policy_type     = "boolean"
-  enforce         = "true"
-  constraint      = "constraints/compute.restrictXpnProjectLienRemoval"
-}
-
-module "org_shared_require_os_login" {
-  source          = "terraform-google-modules/org-policy/google"
-  count           = var.enable_os_login_policy ? 1 : 0
```
daniel-cit, line 96: Thanks for the pull request.

This condition ,`count           = var.enable_os_login_policy ? 1 : 0`, was added because there are some limitations on using OS Login as listed in the 1-org README.md and reproduced bellow:

**Note:** It is possible to enable an organization policy for [OS Login](https://cloud.google.com/compute/docs/oslogin/manage-oslogin-in-an-org) with this module.
OS Login has some [limitations](https://cloud.google.com/compute/docs/instances/managing-instance-access#limitations).
If those limitations do not apply to your workload/environment you can choose to enable the OS Login policy by setting variable `enable_os_login_policy` to `true`.

To be able to maintain this behavior it would be necessary to conditionally add  the OS Login constraint `"constraints/compute.requireOsLogin"` to the map, something like

```
locals {
...
  policies = {
    "constraints/compute.disableNestedVirtualization"   = "boolean",
    "constraints/compute.disableSerialPortAccess"       = "boolean",
    "constraints/compute.disableGuestAttributesAccess"  = "boolean",
    "constraints/compute.vmExternalIpAccess"            = "list",
    "constraints/compute.skipDefaultNetworkCreation"    = "boolean",
    "constraints/compute.restrictXpnProjectLienRemoval" = "boolean",

  }
  os_login_policy = {
    "constraints/compute.requireOsLogin" = "boolean",
  }
  resolved_policies = merge(local.policies, var.enable_os_login_policy ? local.os_login_policy : {})
}

```

daniel-cit (commented):
the org policies:

- org_cloudsql_external_ip_access
- org_disable_sa_key_creation
- org_enforce_bucket_level_access

could also be added to the `policies` map, since only `org_domain_restricted_sharing` uses a different module

1-org/envs/shared/org_policy.tf
bharathkkb: @daniel-cit I was thinking we could just comment it inline with issue and just add a section in the readme instructing users to tweak this part.
```hcl
 policies = {
    "constraints/compute.disableNestedVirtualization"   = "boolean",
    ...
   #  "constraints/compute.requireOsLogin" = "boolean" // disabled due to https://cloud.google.com/compute/docs/instances/managing-instance-access#limitations
    
    }
```
This way, we can also inform them of other policies like https://github.com/terraform-google-modules/terraform-example-foundation/pull/271 without introducing a flag per policy.

daniel-cit: It would also work, it is just important that in the README we list which are the optional ones. 

It will also be necessary to remove the commented org policies from the list in the integration test:

https://github.com/terraform-google-modules/terraform-example-foundation/blob/f87ed1661bec7c2d2c75ea091d1cf5db6ac6943a/test/integration/org/controls/gcloud_policy.rb#L18

1-org/README.md
```diff
@@ -23,7 +23,7 @@ You can change the filters & sinks by modifying the configuration in `envs/share
 
 **Note:** It is possible to enable an organization policy for [OS Login](https://cloud.google.com/compute/docs/oslogin/manage-oslogin-in-an-org) with this module.
 OS Login has some [limitations](https://cloud.google.com/compute/docs/instances/managing-instance-access#limitations).
-If those limitations do not apply to your workload/environment you can choose to enable the OS Login policy by setting variable `enable_os_login_policy` to `true`.
+If those limitations do not apply to your workload/environment you can choose to enable the OS Login policy by uncommenting the "constraints/compute.requireOsLogin" policy constraint in `envs/shared/org_policy.tf` line 28.
```
bharathkkb, line 26:
```suggestion
If those limitations do not apply to your workload/environment you can choose to enable the OS Login policy by uncommenting the "constraints/compute.requireOsLogin" policy constraint in `envs/shared/org_policy.tf` [line 28](https://github.com/terraform-google-modules/terraform-example-foundation/blob/60729c49a53efca8ec37e948acfae8ab5bc4ab44/1-org/envs/shared/org_policy.tf#L28).
```

bharathkkb (approved):
LGTM
/cc @rjerrems

rjerrems:
I am not sure that this convinced this abstraction is necessary, particularly because it makes the common use case of exempting a particular folder or project id from a specific policy more difficult. If a user wanted to exclude a project / folder how would they go about doing that?

IMHO It's also worth keeping in mind, this repo is supposed to be a demonstration of usage of CFT modules so being explicit is encouraged unless there is a strong use case to make things more readable/DRY etc.

bharathkkb:
@rjerrems makes sense and agreed that it is better to be explicit. My original motivation in the issue was to reduce the number of flags we need to expose like `enable_os_login_policy` to satisfy SCC and allow users to easily see a list of all policies. The users could keep the policies that have `exclude_*` out of the list as a separate config.

That being said, now I don't think I have a strong preference either way

rjerrems:
Make sense @bharathkkb - perhaps we can put this on ice for the time being?

Sorry for confusion @meganzhao10 - thanks for the PR!

PR #329, opened 2021-02-05
Clone policies repo once

juliocc:
Currently `tf_plan_validate_all` calls `tf_validate` multiple times, which means the policies repo getting cloned multiple times. This results in an error the second time the clone is attempted since the path already exists

With this PR, `tf-wrapper.sh` verifies the existence of `/workspace/.policies-cloned` before attempting to clone.

### DIFF
--- a/build/tf-wrapper.sh
+++ b/build/tf-wrapper.sh
@@ -138,10 +138,9 @@ tf_validate() {
       cd "$path" || exit
       terraform show -json "${tmp_plan}/${tf_component}-${tf_env}.tfplan" > "${tf_env}.json" || exit 32
       if [[ "$policy_type" == "CLOUDSOURCE" ]]; then
-        if [ ! -f /workspace/.policies-cloned ]; then
-          # only clone the policies repo once
+        # Check if $policy_file_path is empty so we clone the policies repo only once
+        if [ -z "$(ls -A ${policy_file_path} 2> /dev/null)" ]; then
           gcloud source repos clone gcp-policies "${policy_file_path}" --project="${project_id}" || exit 34
-          touch /workspace/.policies-cloned
         fi
       fi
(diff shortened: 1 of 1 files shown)


build/tf-wrapper.sh
```diff
@@ -138,7 +138,11 @@ tf_validate() {
       cd "$path" || exit
       terraform show -json "${tmp_plan}/${tf_component}-${tf_env}.tfplan" > "${tf_env}.json" || exit 32
       if [[ "$policy_type" == "CLOUDSOURCE" ]]; then
-        gcloud source repos clone gcp-policies "${policy_file_path}" --project="${project_id}" || exit 34
+        if [ ! -f /workspace/.policies-cloned ]; then
+          # only clone the policies repo once
+          gcloud source repos clone gcp-policies "${policy_file_path}" --project="${project_id}" || exit 34
+          touch /workspace/.policies-cloned
```
bharathkkb, line 144: Are we using this dir as a marker just for ensuring that the policies have been cloned?
Can't we just clone into a predefined directory `gcloud source repos clone "${policy_file_path}" "${policy_dir}" --project="${project_id}"` and for subsequent runs check for that `policy_dir` exists and is not empty?

juliocc: Yes, /workspace/.policies-cloned is just to check if the repo has been cloned. 

We're already cloning to `$policy_file_path` so I can change line 141 to 
```bash
if [ -n "$(ls -A ${policy_file_path})" ] ; then ...
```
We don't need to check if the directory exists since we already know it does (line 133)

bharathkkb: yup that makes sense to me

bharathkkb (approved):
LGTM
/cc @rjerrems @daniel-cit

Issue #251, opened 2020-09-29
Main README is missing new peering projects


rjerrems:
New peering projects were added in https://github.com/terraform-google-modules/terraform-example-foundation/pull/243

The main README needs to be updated to reflect this.

cc @amandakarina

bharathkkb:
fixed by #336

PR #336, opened 2021-02-10
documentation: Updated main README

chris-ingrisani:
Main README was updated with this PR to reflect new peering projects and close issue #251


bharathkkb (approved):
Thanks for the PR @chris-ingrisani 
The int tests are failing as this is not rebased on develop, but since it's a `readme` change, it's fine as linters passed.

PR #337, opened 2021-02-12
App Infra pipelines

### LINKED ISSUE #316: Deployment pipelines for infra management within service projects
Create an example lightweight deployment pipeline for managing infra resources within service projects provisioned in 4-projects. This will showcase

- operators managing lifecycle of these resources independent to the pipeline used to provision other layers (1-org, 2-envs, 3-networks) .
- best practices for  managing permissions for SAs used for provisioning resources within these projects
- gitops strategies for managing multiple environments

bharathkkb:
fixes #316 
- Adds a common build project per BU in 4-projects
- Accepts a list of apps and per app creates a CSR, 2x pipeline (plan/apply), TF runner image, artifact and state bucket
- single_project which wraps project factory can optionally give cloudbuild_sa role to impersonate the newly created project_sa (with a list of roles as needed)

TODO:
- add 5-app-infra to showcase application infra deployment using this pipeline
- granular roles for configuring CSR/pipelines access


4-projects/modules/infra_pipelines/cloudbuild_builder/Dockerfile
```diff
@@ -0,0 +1,44 @@
+# Copyright 2021 Google LLC
```
rjerrems, line 1: It might be worthwhile thinking about sharing the image/s created in bootstrap, so it is consistent throughout the app pipelines

4-projects/modules/infra_pipelines/variables.tf
```diff
@@ -0,0 +1,75 @@
...
+  type        = string
+  default     = "us-central1"
+}
+
+variable "terraform_apply_branches" {
+  description = "List of git branches configured to run terraform apply Cloud Build trigger. All other branches will run plan by default."
+  type        = list(string)
+
+  default = ["development",
+    "non\\-production", //non-production needs a \ to ensure regex matches correct branches.
```
rjerrems, line 43: I believe we can remove these escape characters from both here and bootstrap. (would need to test, but would be cleaner)

4-projects/modules/infra_pipelines/variables.tf
```diff
@@ -0,0 +1,75 @@
...
+variable "cloudbuild_apply_filename" {
+  description = "Path and name of Cloud Build YAML definition used for terraform apply."
+  type        = string
+  default     = "cloudbuild-tf-apply.yaml"
+}
+
+variable "terraform_version" {
+  description = "Default terraform version."
+  type        = string
+  default     = "0.12.29"
```
rjerrems, line 62: Should we have at least > 0.13 ? (may not be relevant if we share the image)

4-projects/modules/infra_pipelines/variables.tf
```diff
@@ -0,0 +1,75 @@
...
+variable "terraform_version_sha256sum" {
+  description = "sha256sum for default terraform version."
+  type        = string
+  default     = "872245d9c6302b24dc0d98a1e010aef1e4ef60865a2d1f60102c8ad03e9d5a1d"
+}
+
+variable "terraform_validator_release" {
+  description = "Default terraform-validator release."
+  type        = string
+  default     = "2020-09-24"
```
rjerrems, line 74: These values should be consistent with what we are using in bootstrap step if we are going to build per app infra project.

build/cloudbuild-tf-apply.yaml
```diff
@@ -22,8 +22,11 @@ steps:
   args:
   - -c
   - |
-    echo "Setting up gcloud for impersonation"
-    gcloud config set auth/impersonate_service_account ${_TF_SA_EMAIL}
+   tf_sa_email=${_TF_SA_EMAIL}
```
rjerrems, line 25: we could probably just use the substitution directly, rather than setting to a variable first

build/cloudbuild-tf-plan.yaml
```diff
@@ -22,8 +22,11 @@ steps:
   args:
   - -c
   - |
-    echo "Setting up gcloud for impersonation"
-    gcloud config set auth/impersonate_service_account ${_TF_SA_EMAIL}
+    tf_sa_email=${_TF_SA_EMAIL}
+    if [[ -n ${tf_sa_email} ]]; then
```
rjerrems, line 26: same as comment for apply.yaml

build/cloudbuild-tf-apply.yaml
rjerrems, line 25: Also it looks like this substitution won't be available for app infra triggers?

bharathkkb: I was initially trying to use it directly like `if [[ -n ${_TF_SA_EMAIL} ]]; then` , but if `_TF_SA_EMAIL` is not set (in the case of app infra pipelines) cloudbuild fails to start build and complains about  a missing substitution. I tried this as a workaround which seemed to work.

4-projects/modules/infra_pipelines/variables.tf
bharathkkb: in bootstrap step we are not overriding the default set by the module https://github.com/terraform-google-modules/terraform-google-bootstrap/blob/c8db8f3ca09ebb9a99155ac10727f01ec7fd7645/modules/cloudbuild/variables.tf#L146

4-projects/modules/infra_pipelines/variables.tf
bharathkkb: good catch, will update

4-projects/modules/infra_pipelines/cloudbuild_builder/Dockerfile
bharathkkb: ack, I have opened https://github.com/terraform-google-modules/terraform-google-bootstrap/issues/91 to track

4-projects/modules/infra_pipelines/variables.tf
bharathkkb: Won't this be an issue https://github.com/terraform-google-modules/terraform-google-bootstrap/issues/52

build/cloudbuild-tf-apply.yaml
bharathkkb: Correct, that was my reason for wrapping in a if statement

rjerrems:
Overall looking good @bharathkkb, nice work! A few small things to look at, not sure whether we need to build an image per app infra project and could cause some confusion / drift vs the image used for the overall foundation.

4-projects/modules/infra_pipelines/variables.tf
rjerrems: (I think this is a relic of using funkier regexes, rather than the invert option)

bharathkkb:
Thanks for the review @rjerrems 
I have opened an issue https://github.com/terraform-google-modules/terraform-google-bootstrap/issues/91 to store the bootstrap runner image in a GAR which would allow for us to have more restrictive ACLs. Another reason was in case app developers wanted a custom TF image with additional tooling/custom providers preinstalled.

4-projects/modules/infra_pipelines/variables.tf
bharathkkb: updates to 0.13.6

4-projects/modules/infra_pipelines/variables.tf
bharathkkb: I tested and looks like we can remove the escape chars

4-projects/modules/infra_pipelines/variables.tf
rjerrems: I believe we are actually (on the develop branch) https://github.com/terraform-google-modules/terraform-example-foundation/blob/develop/0-bootstrap/main.tf#L131

bharathkkb: good catch, updated

bharathkkb:
@rjerrems I will expose an option to use the foundation runner image in GAR once we have a new bootstrap release

rjerrems (approved):
LGTM - Lets just confirm tests are working prior to merge.

PR #339, opened 2021-02-18
upgrade version for dependant CFT modules

ericyz:


morgante (approved):

1-org/envs/shared/versions.tf
```diff
@@ -17,3 +17,16 @@
 terraform {
   required_version = ">= 0.13"
 }
+terraform {
+  required_version = ">=0.13, <0.14"
```
bharathkkb, line 21: there seems to be conflicting `terraform` blocks with required_version. We should keep these to `required_version >= 0.13` to allow users flexibility

bharathkkb (changes requested):
Thanks for the PR @ericyz

1-org/envs/shared/versions.tf
bharathkkb: this seems to be in other updated versions files as well

1-org/envs/shared/versions.tf
```diff
@@ -17,3 +17,16 @@
 terraform {
```
bharathkkb, line 17: @ericyz we should remove this extraneous `terraform` block here and throughout

bharathkkb (approved):

PR #341, opened 2021-02-19
fix: Always/safely use latest cloudsource policies

deas:
I think there are two issues with cloud source policy repos in `tf-wrapper.sh` we can quickly improve.

- We always want to use the latest version.
- There are scenarios when we fail due to `"${policy_file_path}"` existing from a previous invocation 

```
Step #1 - "tf plan validate all": *************** TERRAFORM VALIDATE ******************
Step #1 - "tf plan validate all":       At environment: envs/production 
Step #1 - "tf plan validate all":       Using policy from: /workspace/policy-library 
Step #1 - "tf plan validate all": *****************************************************
Step #1 - "tf plan validate all": WARNING: This command is using service account impersonation. All API calls will be executed as [org-terraform@cft-seed-xyz.iam.gserviceaccount.com].
Step #1 - "tf plan validate all": WARNING: This command is using service account impersonation. All API calls will be executed as [org-terraform@cft-seed-xyz.iam.gserviceaccount.com].
Step #1 - "tf plan validate all": WARNING: This command is using service account impersonation. All API calls will be executed as [org-terraform@cft-seed-xyz.iam.gserviceaccount.com].
Step #1 - "tf plan validate all": ERROR: (gcloud.source.repos.clone) Directory path specified exists and is not empty
Finished Step #1 - "tf plan validate all"
ERROR
ERROR: build step 1 "gcr.io/cft-cloudbuild-fabc/terraform" failed: step exited with non-zero status: 1
```

### DIFF
--- a/build/tf-wrapper.sh
+++ b/build/tf-wrapper.sh
@@ -138,6 +138,9 @@ tf_validate() {
       cd "$path" || exit
       terraform show -json "${tmp_plan}/${tf_component}-${tf_env}.tfplan" > "${tf_env}.json" || exit 32
       if [[ "$policy_type" == "CLOUDSOURCE" ]]; then
+        if [ -d "${policy_file_path}" ]; then
+          rm -rf "${policy_file_path}" || exit 34
+        fi
         gcloud source repos clone gcp-policies "${policy_file_path}" --project="${project_id}" || exit 34
       fi
       terraform-validator validate "${tf_env}.json" --policy-path="${policy_file_path}" --project="${project_id}" || exit 33


bharathkkb (commented):
Thanks for the PR @deas 
We have a proposed fix in https://github.com/terraform-google-modules/terraform-example-foundation/pull/329 which I think maybe a better solution. We clone only if it has not been cloned during the current run. WDYT?

deas:
@bharathkkb My first approach pretty much looked like #329 . I (probably mistakenly) felt like there where edge cases where the policies may have drifted out of date at execution time. I expect the policy repo to be small, so I chose to make sure and always do a fresh clone.

Even though the approach in #329 may work perfectly fine, the conditional `[ -d "${policy_file_path}" ]` appears a bit confusing to me.

bharathkkb:
> edge cases where the policies may have drifted out of date at execution time

@deas there yes is chance that the during a single run it may be out of date, for instance 

- dev validation passes
- someone changes the policy repo and we do not reclone

However, I think that is acceptable since it only happens within a single run and the subsequent run will clone the fresh repo. @rjerrems any thoughts on this?

>  the conditional [ -d "${policy_file_path}" ] appears a bit confusing to me.

Do you mean the `if [ -z "$(ls -A "${policy_file_path}" 2> /dev/null)" ];` in #329 ? IIUC it is check if the policy path has any files, if no files then clone.

deas:
> > edge cases where the policies may have drifted out of date at execution time
> 
> @deas there yes is chance that the during a single run it may be out of date, for instance
> 
> * dev validation passes
> * someone changes the policy repo and we do not reclone
> 
> However, I think that is acceptable since it only happens within a single run and the subsequent run will clone the fresh repo. @rjerrems any thoughts on this?

FWIW: I am personally ok with a few minutes, but there there should not be edge cases in the order of hours or days. :) (I don't see any now).

> 
> > the conditional [ -d "${policy_file_path}" ] appears a bit confusing to me. 
> 
> Do you mean the `if [ -z "$(ls -A "${policy_file_path}" 2> /dev/null)" ];` in #329 ? IIUC it is check if the policy path has any files, if no files then clone.

Yes, sorry (`[ -d "${policy_file_path}" ]` was my local quick fix).

My point here is that I was wondering "Why don't we just unconditionally clone once and not even worry about it?".

I am fine with #329. ;)

rjerrems:
With regard to "Why don't we just unconditionally clone once and not even worry about it?". I think there is some merit to doing this and removing the logic from the script, with one of the first steps in cloud build YAML config being to clone the repo, which could make things a little simpler. Part of the reason the logic exists at all is because it is in a block of code that gets run multiple times per build.

Overall, my thoughts are:

- Per *build* the repo should only be cloned once (right now this is happening AFAIK, just through logic in bash script)
- Within a build, there may be a number of validations against each environment which I think should use the same consistent version of the policies. Otherwise you could get some strange behavior where one environment uses one commit and a later environment in the same build uses another. (with the trade off it could be N minutes out of date)

rjerrems:
Closing as already fixed https://github.com/terraform-google-modules/terraform-example-foundation/pull/329 - Thanks for highlighting the issue @deas !

PR #343, opened 2021-02-19
Support for hierarchical firewall policies

juliocc:


3-networks/modules/hierarchical_firewall_policy/main.tf
```diff
@@ -0,0 +1,68 @@
...
+  for_each = var.rules
+
+  policy_id               = google_compute_organization_security_policy.policy.id
+  action                  = each.value.action
+  direction               = each.value.direction
+  priority                = each.value.priority
+  target_resources        = each.value.target_resources
+  target_service_accounts = each.value.target_service_accounts
+  enable_logging          = each.value.logging
+  # preview                 = each.value.preview
```
bharathkkb, line 44: nit: remove commented code

3-networks/modules/hierarchical_firewall_policy/main.tf
```diff
@@ -0,0 +1,68 @@
...
+  match {
+    # description = each.value.description
```
bharathkkb, line 46: This looks like it's being used in `envs/shared`

3-networks/modules/base_shared_vpc/firewall.tf
```diff
@@ -69,121 +69,3 @@ resource "google_compute_firewall" "allow_private_api_egress" {
...
 }
-
-
-/******************************************
-  Optional firewall rules
- *****************************************/
-
-// Allow SSH via IAP when using the allow-iap-ssh tag for Linux workloads.
-resource "google_compute_firewall" "allow_iap_ssh" {
-  count   = var.optional_fw_rules_enabled ? 1 : 0
```
bharathkkb, line 80: In shared can we continue to support flags like `optional_fw_rules_enabled`? Perhaps a `merge` on `rules` map for `allow-iap-ssh-rdp` etc?

bharathkkb (commented):
Thanks for the PR @juliocc 
Overall LGTM

3-networks/envs/shared/hierarchical_firewall.tf
```diff
@@ -0,0 +1,102 @@
...
+      direction               = "INGRESS"
+      action                  = "allow"
+      priority                = 5000
+      ranges                  = ["35.235.240.0/20"]
+      ports                   = { "all" = [] }
+      target_service_accounts = null
+      target_resources        = null
+      logging                 = var.firewall_policies_enable_logging
+    }
+    allow-windows-activation = {
```
rjerrems, line 70: without the target tags, how would users make use of these rules? Is the default that this will apply to all instances if no target resource or service accounts are supplied?

3-networks/envs/shared/hierarchical_firewall.tf
```diff
@@ -0,0 +1,102 @@
...
+  name   = "common-firewall-rules"
+  associations = [
+    data.google_active_folder.common.name,
+    data.google_active_folder.bootstrap.name,
+    data.google_active_folder.development.name,
+    data.google_active_folder.production.name,
+    data.google_active_folder.non-production.name,
+  ]
+  rules = {
+    delegate-rfc1918-ingress = {
```
rjerrems, line 29: What are the purpose of these two delegation rules? Is it to allow all internal traffic in both directions prior to VPC firewall rules being evaluated?

3-networks/envs/shared/hierarchical_firewall.tf
```diff
@@ -0,0 +1,102 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+module "hierarchical_firewall_policy" {
+  source = "../../modules/hierarchical_firewall_policy/"
```
rjerrems, line 18: It may be worthwhile considering adding this as a submodule of https://github.com/terraform-google-modules/terraform-google-network

@bharathkkb what do you think?

3-networks/envs/shared/hierarchical_firewall.tf
```diff
@@ -0,0 +1,102 @@
...
+        "10.0.0.0/8",
+        "172.16.0.0/12"
+      ]
+      ports                   = { "all" = [] }
+      target_service_accounts = null
+      target_resources        = null
+      logging                 = false
+    }
+    allow-iap-ssh-rdp = {
+      description             = "Always allow SSH and RDP from IAP"
```
rjerrems, line 60: In the previous configuration, these rules were optional and disabled by default. https://github.com/terraform-google-modules/terraform-example-foundation/blob/develop/3-networks/envs/development/variables.tf#L107

Are we changing that behaviour here?

rjerrems (commented):
Thanks for the PR @juliocc ! I have a few questions about how these changes the current behaviour / user experience & whether or not this should become a module in the network module itself.

3-networks/envs/shared/hierarchical_firewall.tf
bharathkkb: yes I think thats a good idea, this will be a general ask outside of foundations too.

3-networks/envs/shared/hierarchical_firewall.tf
bharathkkb: I believe it's inherited and will be applied to all unless overridden with a regular FW rule. Some patterns here https://cloud.google.com/vpc/docs/firewall-policies-examples#example_4_configure_organization-wide_and_folder-specific_rules

3-networks/modules/hierarchical_firewall_policy/main.tf
juliocc: ACK

3-networks/envs/shared/hierarchical_firewall.tf
juliocc: These rules (`allow-iap-ssh-rdp`, `allow-windows-activation`, `allow-google-hbs-and-hcs`) would always be applied. The only one that can be overridden is `delegate-rfc1918-ingress`. 

Check out section 7.6.1 of the draft document.

3-networks/envs/shared/hierarchical_firewall.tf
juliocc: These two rules delegate RFC1918 traffic to lower levels (i.e. we're not making any decision in regards to that traffic here).

Again, check the document for the rationale.

3-networks/envs/shared/hierarchical_firewall.tf
juliocc: I can take care of that.

3-networks/envs/shared/hierarchical_firewall.tf
juliocc: Indeed. Before, SSH/RDP from IAP was optional, whereas now is always enabled.

juliocc:
@rjerrems agreed that a module would be better. I fact, some of the code actually comes from the [Hierarchical Firewall module](https://github.com/terraform-google-modules/cloud-foundation-fabric/tree/master/modules/folder#hierarchical-firewall-policies) of [Cloud Foundation Fabric](https://github.com/terraform-google-modules/cloud-foundation-fabric)

I think it should be fairly easy to port some of the code back to CFT. Let me know if we want to do that now or if we want to merge this as is.

3-networks/modules/base_shared_vpc/firewall.tf
juliocc: We can do that but it means all environments get the same policy (i.e. can't have an environment without the optional rules and another environment with them)

What do you think?

bharathkkb: I see, in that case perhaps these should be regular FW rules which we can control per env and can override inherited HFW rules. @rjerrems WDYT?

rjerrems: I think that makes sense to me

rjerrems:
> @rjerrems agreed that a module would be better. I fact, some of the code actually comes from the [Hierarchical Firewall module](https://github.com/terraform-google-modules/cloud-foundation-fabric/tree/master/modules/folder#hierarchical-firewall-policies) of [Cloud Foundation Fabric](https://github.com/terraform-google-modules/cloud-foundation-fabric)
> 
> I think it should be fairly easy to port some of the code back to CFT. Let me know if we want to do that now or if we want to merge this as is.

I think we can probably merge once we agree on implementation, then have a follow up task to implement as a submodule in the CFT network module & replace the reference here. @bharathkkb WDYT?

3-networks/envs/shared/hierarchical_firewall.tf
rjerrems: Ok cool - just clarifying that is intended with the team as in the previous release we explicitly did the opposite.

3-networks/envs/shared/hierarchical_firewall.tf
rjerrems: Ok thanks - just read through it. The doc in 7.6.1 also mentions blocking egress to the internet but I can't see a rule for that here?

rjerrems (approved):
@juliocc we should be good to go ahead with this implementation based on consensus with the team. Can you please rebase from develop as there are now conflicts?

juliocc:
ACK. I'll try to get it done today

juliocc:
Rebased and all checks pass. Let me know if you want me to merge.

rjerrems:
great thanks @juliocc

PR #345, opened 2021-02-21
implement support for Partner Interconnect

gaspar-chilingarov:
- add 3-networks/modules/partner_interconnect module
- update documentation
- provide partner_interconnect.tf.example modules for each environment,
  including hub_and_spoke mode.


3-networks/envs/production/interconnect.auto.tfvars.example
```diff
@@ -0,0 +1,4 @@
+
+enable_partner_interconnect = true
```
rjerrems, line 2: Should these be default values for the variables, rather than something that needs to be tfvars?

rjerrems (approved):
LGTM - Only one minor nit about tfvars files.

3-networks/envs/production/interconnect.auto.tfvars.example
gaspar-chilingarov: default is false :P so no Partnet IC by default :) just renaming parnet_interconnect.tf.example is not enough, as we need value changes in primary main.tf.

rjerrems: Ok I understand better now. Why not just set these values when you call the partner interconnect module? (in example files) - the less tfvars files that need to be managed, the better IMHO and you can customise these per environment in the actual code if you want.

Also FYI, for the other tfvars files they have been sym linked to this folder so you only need to update them once. https://github.com/gaspar-chilingarov/terraform-example-foundation/tree/feature/add-partner-interconnect-example/3-networks

gaspar-chilingarov: > Why not just set these values when you call the partner interconnect module? (in example files) 

Because I need to feed [different ASN value in the main.tf depending if this needs to support patner interconnect or not](https://github.com/terraform-google-modules/terraform-example-foundation/pull/345/files#diff-bd2cb2e885c48da5270bd9b2b29f4e6933ab5ab2d4ac4804179031725817847e). I cannot set variables from the .tf file. At least no non-hacky ways I'm aware of.

> other tfvars files they have been sym linked to this folder
Well, interconnect configuration (or even presence/absense) cound be different per environment. 

Also there is a reason why we don't do it with .example files - as the target of the symlink should be always in place. (so user must rename file in the top directory and then also symlink in all environments)

so straightforward solution is to decide not to use `.example` as we do now, move all of them into normal .tf files and add `count = enabled_{interconnect|partner|vpn}` condition. In that way we have don't have variance in TF code which is executed, just some modules/resources will not be instantiated.

rjerrems: Yeah I think we can get rid of the example files and just have a condition now, because we will be on terraform 0.13+ - so I agree that is a good idea.

Would you like to do this in a follow up PR?

@bharathkkb FYI

bharathkkb: sgtm

> add `count = enabled_{interconnect|partner|vpn}` condition

maybe instead of bool flags we can condense them into a single string input and switch base on string content `interconnect|partner|vpn`

gaspar-chilingarov: well, if we want to integrate VPN/Dedicated (need rework anyway, as configurations are wrong)/Partner - sure, we can do that in separate PR

PR #347, opened 2021-02-23
Initial Naming Convention PR

JLangevine:
Updates for DNS Hub


rjerrems:
hi @JLangevine - is this PR still active? It looks like there is some work to update a number of other things, right now the only changes are the single data resource which will probably break other things without actually changing the project name?

Thanks

PR #372, opened 2021-03-09
feat/external-ip

chris-ingrisani:
This PR is to add an external IP that can be used for HTTP load balancing.


bharathkkb:
Hi @chris-ingrisani 
This change should be made to @jasonbisson fork for example-foundation

PR #393, opened 2021-03-25
update bootstrap example

z-shah:
- Update `project_id` and `state_bucket_name` in cloudbuild and seed module examples to follow whitepaper standards

### DIFF
--- a/0-bootstrap/main.tf
+++ b/0-bootstrap/main.tf
@@ -48,7 +48,6 @@ module "seed_bootstrap" {
   sa_enable_impersonation        = true
   parent_folder                  = var.parent_folder == "" ? "" : local.parent
   org_admins_org_iam_permissions = local.org_admins_org_iam_permissions
-  project_prefix                 = var.project_prefix
 
   project_labels = {
     environment       = "bootstrap"
@@ -119,7 +118,6 @@ module "cloudbuild_bootstrap" {
   sa_enable_impersonation     = true
   cloudbuild_plan_filename    = "cloudbuild-tf-plan.yaml"
   cloudbuild_apply_filename   = "cloudbuild-tf-apply.yaml"
-  project_prefix              = var.project_prefix
   cloud_source_repos          = var.cloud_source_repos
   terraform_validator_release = "2021-01-21"
   terraform_version           = "0.13.5"
--- a/0-bootstrap/main.tf
+++ b/0-bootstrap/main.tf
@@ -38,7 +38,7 @@ module "seed_bootstrap" {
   version                        = "~> 2.1"
   org_id                         = var.org_id
   folder_id                      = google_folder.bootstrap.id
-  project_id                     = "prj-example"
+  project_id                     = "prj-c-seed"
   state_bucket_name              = "bkt-b-tfstate-${random_id.suffix.hex}"
   billing_account                = var.billing_account
   group_org_admins               = var.group_org_admins
@@ -109,7 +109,7 @@ module "cloudbuild_bootstrap" {
   version                     = "~> 2.1"
   org_id                      = var.org_id
   folder_id                   = google_folder.bootstrap.id
-  project_id                  = "prj-cicd"
+  project_id                  = "prj-c-cicd"
   billing_account             = var.billing_account
   group_org_admins            = var.group_org_admins
   default_region              = var.default_region
--- a/0-bootstrap/main.tf
+++ b/0-bootstrap/main.tf
@@ -29,11 +29,17 @@ resource "google_folder" "bootstrap" {
   parent       = local.parent
 }
 
+resource "random_id" "suffix" {
+  byte_length = 2
+}
+
 module "seed_bootstrap" {
   source                         = "terraform-google-modules/bootstrap/google"
   version                        = "~> 2.1"
   org_id                         = var.org_id
   folder_id                      = google_folder.bootstrap.id
+  project_id                     = "prj-example"
+  state_bucket_name              = "bkt-b-tfstate-${random_id.suffix.hex}"
   billing_account                = var.billing_account
   group_org_admins               = var.group_org_admins
   group_billing_admins           = var.group_billing_admins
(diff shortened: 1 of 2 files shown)


0-bootstrap/main.tf
```diff
@@ -29,11 +29,17 @@ resource "google_folder" "bootstrap" {
...
+resource "random_id" "suffix" {
+  byte_length = 2
+}
+
 module "seed_bootstrap" {
   source                         = "terraform-google-modules/bootstrap/google"
   version                        = "~> 2.1"
   org_id                         = var.org_id
   folder_id                      = google_folder.bootstrap.id
+  project_id                     = "prj-example"
```
rjerrems, line 41: I believe the name should be something like this:

`prj-c-seed`

With the c standing for common, another example here:

https://github.com/terraform-google-modules/terraform-example-foundation/blob/master/1-org/envs/shared/projects.tf#L27

And more detail of naming conventions here on page 9 https://services.google.com/fh/files/misc/google-cloud-security-foundations-guide.pdf

0-bootstrap/main.tf
```diff
@@ -103,6 +109,7 @@ module "cloudbuild_bootstrap" {
   version                     = "~> 2.1"
   org_id                      = var.org_id
   folder_id                   = google_folder.bootstrap.id
+  project_id                  = "prj-cicd"
```
rjerrems, line 112: similar to above, this should be `prj-c-cicd`

0-bootstrap/main.tf
```diff
@@ -29,11 +29,17 @@ resource "google_folder" "bootstrap" {
...
+  byte_length = 2
+}
+
 module "seed_bootstrap" {
   source                         = "terraform-google-modules/bootstrap/google"
   version                        = "~> 2.1"
   org_id                         = var.org_id
   folder_id                      = google_folder.bootstrap.id
+  project_id                     = "prj-example"
+  state_bucket_name              = "bkt-b-tfstate-${random_id.suffix.hex}"
```
rjerrems, line 42: @bharathkkb - we need to add the random suffix here to ensure builds don't fail / good UX if users don't change defaults. Do you think it might be worth adding a flag to bootstrap module `random_suffix` to toggle if users want that for project / bucket and other resources? (we could do this separately and then remove this later)

rjerrems (changes requested):
Thanks @z-shah - there are a couple of changes required to the naming conventions used. Let me know if it doesn't make sense.

z-shah:
Thanks @rjerrems. I have made changes based on your comments

rjerrems:
G'day @z-shah - it looks like tests are failing because they are still referencing the project prefix. https://github.com/terraform-google-modules/terraform-example-foundation/blob/develop/test/fixtures/bootstrap/main.tf#L25

0-bootstrap/main.tf
bharathkkb: @rjerrems that sgtm as a follow up

0-bootstrap/main.tf
daniel-cit: @rjerrems Considering the naming examples in the guide and the reason we added the `var.project_prefix` to prevent namespace collision, shouldn't the project be named as:

```terraform
"${var.project_prefix}-b-seed-${random_id.suffix.hex}"
```

0-bootstrap/main.tf
daniel-cit: @rjerrems Considering the naming examples in the guide and the reason we added the `var.project_prefix` to prevent namespace collision, shouldn't the project be named as:

```terraform
"${var.project_prefix}-b-cicd-${random_id.suffix.hex}"

File: 0-bootstrap/main.tf (hunk 3 above)
Comment (daniel-cit) [reply 0-bootstrap/main.tf:42]: @rjerrems  Considering the naming examples in the guide and the reason we added the `var.project_prefix` to prevent namespace collision, shouldn't the bucket be named as:

```terraform
"bkt-${project_prefix}-b-seed-tfstate-${random_id.suffix.hex}"
```

0-bootstrap/main.tf
z-shah: @rjerrems  
 @daniel-cit raises a valid point here. I suspect not using `var.project_prefix` in the resource name can cause other conflicts.
 
Here's build logs from the previous commit that failed as well.
https://pantheon.corp.google.com/cloud-build/builds;region=global/78cd2aac-7927-4730-a9d0-5660ad4fd0b9;step=12?project=cloud-foundation-cicd

```
         91: resource "google_dns_policy" "default_policy" {
         on ../../../3-networks/envs/shared/dns-hub.tf line 91, in resource "google_dns_policy" "default_policy":
       
       Error: Error creating Policy: googleapi: Error 409: The resource 'entity.policy' named 'dp-dns-hub-default-policy' already exists, alreadyExists
```

rjerrems: Yeah thanks for pickup @daniel-cit - I think we might need to add this back @z-shah . I discussed with @bharathkkb last week and we think that we should retain the project_prefix for now and we will try to make this more readable in a future release.

0-bootstrap/main.tf
rjerrems: The name that would generate would not be the best, it would expand to something like "bkt-prj-b-seed-tfstate-237sd" if the user does not change the defaults, but I suppose it is closer to the guide with the `bkt` prefix at the beginning

daniel-cit: In the guide the description is:

```
bkt-project-name{-bucket-label}
```
and the example is

```
bkt-p-acde-shared-base-data
```

one problem we can have if the prefix is not used is that the buckets from different projects would fall in to the same similar patterrn:

`prj-b-seed-<random_1>`  **=>** `bkt-b-seed-tfstate-<random_2>`
and 
`tst-b-seed-<random_3>`  **=>**  `bkt-b-seed-tfstate-<random_4>`

since there is only one namespace for bucket names it would increase the chance of a collision.

z-shah:
closing this PR. I'll create a new one based on suggestions and to avoid merge conflicts with develop branch

PR #394, opened 2021-03-26
Feat/infra-tests-1

chris-ingrisani:
This is the first PR to check for roles/editor permissions on the new base project SAs created by the infra pipelines in issue #337

### DIFF
--- a/test/integration/projects/controls/gcloud_projects.rb
+++ b/test/integration/projects/controls/gcloud_projects.rb
@@ -171,6 +171,7 @@ control 'gcloud-projects' do
           end
         end
       end
+
       describe command("gcloud iam service-accounts get-iam-policy #{base_project_sa[environment_code][business_unit]} --format=json") do
         its(:exit_status) { should eq 0 }
         its(:stderr) { should eq '' }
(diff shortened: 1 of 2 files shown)


test/integration/projects/controls/gcloud_projects.rb
```diff
@@ -196,6 +197,31 @@
...
+
+        describe "#{base_projects_id[environment_code][business_unit]} IAM Policy" do
+          it 'should exist' do
+            expect(data).to_not be_empty
+          end
+          it "#{base_project_sa[environment_code][business_unit]} Base project SA should have the Editor role on #{base_projects_id[environment_code][business_unit]}" do
+            expect(data['bindings'][1]['members']).to include(
+              "serviceAccount:#{base_project_sa[environment_code][business_unit]}"
+            )
+            expect(data['bindings'][1]['role']).to eq "roles/editor"
```
bharathkkb, line 220: What is `['bindings'][0]['members']` member? I am wondering since this is a project IAM if we were to add other members this index based access could break. Could we do something like 
```rb
expect(data).to include (
  members: including("serviceAccount:#{base_project_sa[environment_code][business_unit]}"),
  role: "roles/editor"
)
```

bharathkkb (commented):
Thanks for the PR @chris-ingrisani

test/integration/projects/controls/gcloud_projects.rb
chris-ingrisani: That makes sense to me - I was trying to follow the structure of the other test that was already created.  Do you want me to change both or keep the first check as is and change only the project IAM check that I recently added?

bharathkkb: Would be great if you can change both, but with service account IAM, I am not as concerned because it is only at the service account level.

bharathkkb:
we recently merged develop into main, please switch base to main branch and rebase

PR #397, opened 2021-03-29
Update terraform-validator version, instructions and CMEK bucket

daniel-cit:
This PR:

- Updates the version of terraform-validator used to version `2021-03-22`
- Update the README instructions regarding manual execution of terraform-validator

This PR also replaces the module used in the creation of the Google Cloud Storage bucket in the CMEK examples in the 4-projects environments.

From the [main module](https://github.com/terraform-google-modules/terraform-google-cloud-storage) 
https://github.com/terraform-google-modules/terraform-example-foundation/blob/137a379e8ec31dfe7e01d7de58b9b8f050259384/4-projects/business_unit_1/production/example_storage_cmek.tf#L69
to the [simple bucket](https://github.com/terraform-google-modules/terraform-google-cloud-storage/tree/master/modules/simple_bucket) module

```terraform
   source             = "terraform-google-modules/cloud-storage/google//modules/simple_bucket"
```

This is necessary because the way the main module [construct](https://github.com/terraform-google-modules/terraform-google-cloud-storage/blob/master/main.tf) the bucket attribute in the context of a random bucket name

```
locals {
...
  names_set    = toset(var.names)
...
resource "google_storage_bucket" "buckets" {
  for_each = local.names_set
...
  uniform_bucket_level_access = lookup(
    var.bucket_policy_only,
    lower(each.value),
    true,
  )

```

makes the after state of the bucket to be known only after apply which make terraform-validator [crash](https://github.com/GoogleCloudPlatform/terraform-validator/issues/201). 

Running the terraform-validator version in the master branch
which has the [bugfix for the crash](https://github.com/GoogleCloudPlatform/terraform-validator/commit/3845101d519c5f8b866340f67dde6184e8c83000), terraform-validator rises a false positive.

```
Found Violations:

Constraint GCPStorageBucketPolicyOnlyConstraintV1.require_bucket_policy_only on resource //storage.googleapis.com/placeholder-fgDsc2WD: //storage.googleapis.com/placeholder-fgDsc2WD does not have bucket policy only enabled.
``` 

both issues, the crash and the false positive do not happen when using the [simple bucket](https://github.com/terraform-google-modules/terraform-google-cloud-storage/tree/master/modules/simple_bucket) module

This change can be made because the CMEK creates a single bucket.


0-bootstrap/README.md
```diff
@@ -37,10 +37,10 @@ If you are using the `jenkins_bootstrap` sub-module, please see [README-Jenkins]
...
-1. To run terraform-validator steps please follow these [instructions](https://github.com/forseti-security/policy-library/blob/master/docs/user_guide.md#install-terraform-validator) and install the latest version.
+1. To run terraform-validator steps please follow the [instructions](https://github.com/forseti-security/policy-library/blob/master/docs/user_guide.md#install-terraform-validator) in the **Install Terraform Validator** section and install version `2021-03-22`. You will also need to rename the binary from `terraform-validator-<your-platform>` to `terraform-validator`.
     1. Run `terraform plan -input=false -out bootstrap.tfplan`
     1. Run `terraform show -json bootstrap.tfplan > bootstrap.json`
-    1. Run `terraform-validator validate bootstrap.json --policy-path="../policy-library" --project <A-VALID-PROJECT>` and check for violations.
+    1. Run `terraform-validator validate bootstrap.json --policy-path="../policy-library" --project <A-VALID-PROJECT-ID>` and check for violations (`<A-VALID-PROJECT-ID>` must be an existing project you have access to, this is necessary because Terraform-validator needs to bind resources to a valid Google Cloud Platform project).
```
bharathkkb, line 43: > to bind resources to

IIRC we use the project for parsing into CAI format? `binding` may give the wrong impression

bharathkkb (approved):

0-bootstrap/README.md
daniel-cit: Replaced with `link`

PR #450, opened 2021-04-23
fix: add infra pipeline CB SA role test

chris-ingrisani:
This PR adds the test for the editor role on the Cloud Build service account and also removes the indexing from both service account tests.  Rebased on main according to PR #394.


chris-ingrisani:
@bharathkkb I couldn't get the test environment to spin up today, I kept getting billing errors, but this was the proposed change: 
`data["bindings"].to include(...)`.  Indexing appears to work consistently, but I could try running it once with the new proposed change.

bharathkkb:
@chris-ingrisani I think this should be fine for now, we can refactor if it breaks

bharathkkb (approved):

Issue #442, opened 2021-04-21
3-networks deployment feedback


daniel-cit:
Feedback from the deploy of step `3-networks`:

- [x] There is no information regarding the `enable_hub_and_spoke` and `enable_hub_and_spoke_transitivity` configuration in the main README.md.
- [x]  `enable_hub_and_spoke` and `enable_hub_and_spoke_transitivity` should be included commented in the https://github.com/terraform-google-modules/terraform-example-foundation/blob/master/3-networks/common.auto.example.tfvars file, otherwise the user will not know where to add them.
- [x] The main README should have a section with information regarding the two possible network models, at least a reference to the relevant section in the guide.
- [x] the configuration of the variable `target_name_server_addresses` in the file `shared.auto.example.tfvars` lacks context for their usage in the foundation
- [x] there is no context for the variable `domain` in the `common.auto.example.tfvars` file. It is not clear that it is the DNS name of a peering managed zone unless you go to one of the READMEs inside one of the environment in the `envs` folder. It is not evidente that its value **must** end with a period. 
- [x] The example for the terraform service account should be more helpful, instead of 
https://github.com/terraform-google-modules/terraform-example-foundation/blob/1d5b05bf35e98d62b4145a349d490df872449dee/3-networks/common.auto.example.tfvars#L19 it should be:
```terraform
terraform_service_account = "org-terraform@prj-b-seed-1234.iam.gserviceaccount.com"
```
- [ ] The terraform service account created in step `0-bootstrap` expects to be the owner of the project as a side effect of creating the projects, since not all necessary permissions were added to the terraform account when it was created. for example the service account is missing the necessary permission to create firewall rules. This was found out when using the workaround for the quota limit issue from https://github.com/terraform-google-modules/terraform-example-foundation/issues/417
- [x] In the deploy instructions with Cloud build, the user should be instructed to wait for the cloud build for the `development` branch to finish before starting the `non-production` build to avoid concurrent peering changes
```
Step #4 - "tf apply": Error: Error adding network peering: googleapi: Error 400: There is a peering operation in progress on the local or peer network. Try again later., badRequest
Step #4 - "tf apply": 
Step #4 - "tf apply":   on .terraform/modules/restricted_shared_vpc.peering/modules/network-peering/main.tf line 22, in resource "google_compute_network_peering" "local_network_peering":
Step #4 - "tf apply":   22: resource "google_compute_network_peering" "local_network_peering" {
```
[...]

PR #457, opened 2021-04-26
Fix 3-networks uat issues

daniel-cit:
This pull request fixes part of the issue from https://github.com/terraform-google-modules/terraform-example-foundation/issues/442

A fix for missing roles for the org-terraform service account will be part of  another PR.

### DIFF
--- a/3-networks/shared.auto.example.tfvars
+++ b/3-networks/shared.auto.example.tfvars
@@ -14,4 +14,6 @@
  * limitations under the License.
  */
 
+// List of IPv4 address of target name servers for the forwarding zone configuration.
+// Cloud DNS will select the best available name server.
 target_name_server_addresses = ["192.168.0.1", "192.168.0.2"]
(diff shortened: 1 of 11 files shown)


3-networks/shared.auto.example.tfvars
```diff
@@ -14,4 +14,6 @@
  * limitations under the License.
  */
 
+// List of IPv4 address of target name servers for the forwarding zone configuration.
+// Cloud DNS will select the best available name server.
```
rjerrems, line 18: Can you elaborate what you mean here in terms of Cloud DNS selecting the best available name server?

rjerrems (approved):
LGTM

/cc @bharathkkb

3-networks/shared.auto.example.tfvars
daniel-cit: Hi @rjerrems this was an extract from the dns.googleapis.com documentation for forwardingConfig.targetNameServers 
https://cloud.google.com/config-connector/docs/reference/resource-docs/dns/dnsmanagedzone#spec

I replaced it with a link to the [Cloud DNS forwarding zones](https://cloud.google.com/dns/docs/overview#dns-forwarding-zones) documentation that will better explain what these servers are used for.

bharathkkb (approved):

Issue #444, opened 2021-04-21
4-projects deployment feedback


daniel-cit:
Feedback from the deploy of step `4-projects `

- [x] Step 7 needs to reference where users can check possible variables (i.e: 3-networks' README asks the user to refer to the document containing all possible variables).
- [x] It is not clear in this step what is the variable to turn on the hub and spoke and where to add it as there are 8 `auto.example.tfvars` files. 
- [x] For example, if we added it to the common.auto.tfvars file we got the Warning below:
```
Warning: Value for undeclared variable
 
The root module does not declare a variable named "enable_hub_and_spoke" but a
value was found in file "common.auto.tfvars". To use this value, add a
"variable" block to the configuration.
```
- [x] Step 18: Same comments as in 3-networks step 13 (I believe it would be better to create and apply the changes first to the non production environments. This may provide the user the chance to fix outside of production any issues that can occur during the process).

bharathkkb:
https://github.com/terraform-google-modules/terraform-example-foundation/pull/436#discussion_r618678933 todo

PR #459, opened 2021-04-26
fix 4-projects uat feedback issues

daniel-cit:
Fixes https://github.com/terraform-google-modules/terraform-example-foundation/issues/444

### DIFF
--- a/4-projects/development.auto.example.tfvars
+++ b/4-projects/development.auto.example.tfvars
@@ -15,3 +15,5 @@
  */
 
 perimeter_name = "sp_d_shared_restricted_default_perimeter_????"
+
+//enable_hub_and_spoke = true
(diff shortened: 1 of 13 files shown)


4-projects/development.auto.example.tfvars
```diff
@@ -15,3 +15,5 @@
  */
 
 perimeter_name = "sp_d_shared_restricted_default_perimeter_????"
+
+//enable_hub_and_spoke = true
```
rjerrems, line 19: Should this not go in the common tfvars file so the user does not have to set it three times?

rjerrems (approved):
LGTM - one minor nit around tfvars.

/cc @bharathkkb

4-projects/development.auto.example.tfvars
daniel-cit: @rjerrems If we add it in the commons tfvars files we get a warning that we are setting a variable that does not exist when running the shared envs for bu1 and bu2 .

rjerrems: aha got it, thanks!

Issue #445, opened 2021-04-21
5-app-infra deployment feedback


daniel-cit:
Feedback from the deploy of step `5-app-infra`:

- [ ] Step 1: The instruction “run terraform output cloudbuild_project_id in the `4-projects/business_unit_1/shared folder`” does not work if the previous step was made using Cloud Build, the correct path should be `gcp-projects/business_unit_1/shared`
- [ ] Step 1: This same step from `1-org` clones a repo with the same name as the one in this step, `gcp-policies`. Would be good to make it clear to the user that they are not the same.
- [ ] Step 15: It is not clear what value should be used in the project_service_account variable. Up to now we have used Terraform SA and Cloud Build SAs along all Terraforms. There should be a reference to a README explaining the variables.	
- [ ] Step 19-22: Would be good to make clear that the `YOUR_CLOUD_BUILD_PROJECT_ID` mentioned on the URL is the one we got in the first step of the current README (step 5) because all the previous steps we used the project that was created on `0-bootstrap`.

PR #460, opened 2021-04-26
fix 5-app-infra uat feedback issues

daniel-cit:
Fixes https://github.com/terraform-google-modules/terraform-example-foundation/issues/445

### DIFF
--- a/5-app-infra/README.md
+++ b/5-app-infra/README.md
@@ -24,7 +24,7 @@ organizational policies.</td>
 </tr>
 <tr>
 <td><a
-href="https://github.com/terraform-google-modules/terraform-example-foundation/tree/master/2-environments">2-environments</a></td>
+href="https://github.com/terraform-google-modules/terraform-example-foundation/tree/master/2-environments"><span style="white-space: nowrap;">2-environments</span></a></td>
 <td>Sets up development, non-production, and production environments within the
 Google Cloud organization that you've created.</td>
 </tr>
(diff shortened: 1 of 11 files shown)


5-app-infra/README.md
```diff
@@ -75,14 +76,14 @@ This Compute Engine instance is created using the base network from step `3-netw
 
 ### Deploying with Cloud Build
 
-1. Get the project ID from the previous procedure. Run the following in the
-   `4-projects/business_unit_1/shared` folder.
+1. Get the Infra Pipeline project ID from the previous step. Run the following in the
+   `gcp-projects/business_unit_1/shared` folder.
```
rjerrems, line 80: This makes the assumption you are running terraform locally/have it setup.

5-app-infra/README.md
```diff
@@ -75,14 +76,14 @@ This Compute Engine instance is created using the base network from step `3-netw
...
    ```
    terraform output cloudbuild_project_id
    ```
-1. Clone repo.
+1. Clone the policies repo. (This repo has the same name of the repo created in step 1-org but it is from a different project, you may need to rename the previous one to prevent a collision if you are cloning it in the same folder).
```
rjerrems, line 84: Why aren't we using the same policy repo as the rest of the foundation? Is this so app pipeline can have a different set of rules?

rjerrems (approved):
LGTM - A few comments regarding the setup, but these are not related to your changes.

/cc @bharathkkb

5-app-infra/README.md
daniel-cit: Yes, because step in step 4-projects we tell the user that they have to run manually 

- gcp-projects/business_unit_1/shared
- gcp-projects/business_unit_2/shared

Because development, non-production, and production depends on the result of the shared folder, the same dependency with have in step 3-networks regarding the DNS hub

5-app-infra/README.md
daniel-cit: @rjerrems We are using the same policy files as the rest of the foundation, but not the same policy repository.

The infra pipeline has its own repositories, on for the code and one for the policies in the infra-pipeline project.

They can have different values but they start the same.

the infra-pipeline is independent from the main pipeline

daniel-cit: there is a issue to let the user use the artifact image created in 0-bootstrap.
we could open an issue to allow reuse of the policy repo too.

rjerrems: Yeah I think we should do that, it will be a confusing customer experience to have to push the same policies to two different repos and create a greater opportunity for human error.

PR #463, opened 2021-04-27
Update google cloud sdk min version to 319.0.0 and use GA version of gcloud scc notifications

daniel-cit:
This PR:
- Update google cloud sdk min version to 319.0.0 (the same version of the docker image for developer-tools:0.13  )
- Use GA version of `gcloud scc notifications`


rjerrems (approved):
LGTM

/cc @bharathkkb

PR #479, opened 2021-05-14
chore: add blueprint attribution

### LINKED ISSUE #473: Add module attribution
We should make sure we have attribution on this repo.

bharathkkb:
fixes #473 

I also moved all min versions constraints into respective inner modules instead of root modules as most of our root modules are compositions of these modules across envs.


0-bootstrap/versions.tf
```diff
@@ -31,4 +31,9 @@ terraform {
       version = "~> 2.3"
     }
   }
+
+  provider_meta "google" {
+    module_name = "blueprints/terraform/terraform-example-foundation:bootstrap/v1.0.0"
```
rjerrems, line 36: should this now be v2.0.0?

bharathkkb: @rjerrems release-please should autoupdate this

morgante (approved):

PR #480, opened 2021-05-14
add DNS zone for artifact registry

daniel-cit:
This Pull Request adds a DNS zone for `pkg.dev` in the base and restricted networks to allow usage of Google Artifact Registry.


bharathkkb (approved):
Thanks for the fix @daniel-cit

PR #490, opened 2021-06-17
Update terraform to 0.13.7

vovinacci:
This PR sets the `terraform` version to `0.13.7` in the code and and in README references.

Main reasoning behind this PR is that `0.13.7` is the latest version (and most likely the last one in `0.13`) and it's not easy to get 0.13.6 as a package. So, it helps with local development a lot as it won't require to resolve `state snapshot was created by Terraform v0.13.7, which is newer than current v0.13.6` error.


vovinacci:
@daniel-cit @bharathkkb @rjerrems Integration tests failed, could you please advise what went wrong as I don't have access to Cloud Build project. Thanks and much appreciated!

rjerrems:
@vovinacci - I have taken a look at it looks like one of the steps is failing (across a number of PRs). Worth noting that the test runner is currently on 0.13.5 - not sure if we have plans to update @bharathkkb ?

bharathkkb:
@rjerrems `0.13.7` was a backport due to https://discuss.hashicorp.com/t/terraform-updates-for-hcsec-2021-12/23570 and we have not cut a backport test container image as we were already at `0.15.x`.

I don't expect the runner to complain since we should only have >0.13 constraints in config.

vovinacci:
@rjerrems Many thanks for the fix, all checks are back to green.

@bharathkkb Having `0.15` would be very nice. Any plans to add it as a default? I may imagine that some modules may not support it yet, but you may know all the details better that anyone else :)

rjerrems (approved):
LGTM

bharathkkb:
@vovinacci we are hoping to update all modules upstream to support 1.0 soon so this would also support 1.0. We would not require 1.0 in requirements but will switch our CI to do testing with 1.0. I expect current release to be mostly compatible with minor changes (deprecated list() fn, sensitive values being set)

PR #491, opened 2021-06-17
Remove redundant header

Walliee:


Walliee:
Given the bold text in the following line, header seems unnecessary.
image

rjerrems (approved):
LGTM - Thanks for picking this up.

Issue #494, opened 2021-06-28
error creating project prj-c-dns-hub-c98c (prj-c-dns-hub): googleapi: Error 409: Requested entity already exists, alreadyExists.


akorzynski:
Hello! 

I'm creating the infrastructure using the instructions in the terraform-example-foundation repo. The process failed due to project and billing quotas, as predicted in the Troubleshooting file in the repo: 
[Project quota exceeded](https://github.com/terraform-google-modules/terraform-example-foundation/blob/master/docs/TROUBLESHOOTING.md#project-quota-exceeded)
[Billing quota exceeded](https://github.com/terraform-google-modules/terraform-example-foundation/blob/master/docs/TROUBLESHOOTING.md#Billing-quota-exceeded)
After requesting Google to increase the quotas, Terraform fails complaining that the projects already exist:

Step #4 - "tf apply": Error: error creating project prj-c-dns-hub-c98c (prj-c-dns-hub): googleapi: Error 409: Requested entity already exists, alreadyExists. If you received a 403 error, make sure you have the `roles/resourcemanager.projectCreator` permission
Step #4 - "tf apply": 
Step #4 - "tf apply":   on .terraform/modules/dns_hub/modules/core_project_factory/main.tf line 65, in resource "google_project" "main":
Step #4 - "tf apply":   65: resource "google_project" "main" {
Step #4 - "tf apply": 
Step #4 - "tf apply": 
Step #4 - "tf apply": 
Step #4 - "tf apply": Error: error creating project prj-c-interconnect-8fd0 (prj-c-interconnect): googleapi: Error 409: Requested entity already exists, alreadyExists. If you received a 403 error, make sure you have the `roles/resourcemanager.projectCreator` permission
Step #4 - "tf apply": 
Step #4 - "tf apply":   on .terraform/modules/interconnect/modules/core_project_factory/main.tf line 65, in resource "google_project" "main":
Step #4 - "tf apply":   65: resource "google_project" "main" {
Step #4 - "tf apply": 
Step #4 - "tf apply": 
Step #4 - "tf apply": 
Step #4 - "tf apply": Error: error creating project prj-c-logging-9a94 (prj-c-logging): googleapi: Error 409: Requested entity already exists, alreadyExists. If you received a 403 error, make sure you have the `roles/resourcemanager.projectCreator` permission
Step #4 - "tf apply": 
[...]

Issue #378, opened 2021-03-17
 googleapi: Error 400: Precondition check failed., failedPrecondition


louiskimlevu:
Hello,

I am receiving this error when running tf apply for the part 2 (environments). I saw this issue can come from limits of the free trial but I already enabled the standard subscription in my organization.

```
Error: Error setting billing account "011EE1-C53D47-660F3B" for project "projects/prj-d-shared-base-9561": googleapi: Error 400: Precondition check failed., failedPrecondition

  on .terraform/modules/env.base_shared_vpc_host_project/modules/core_project_factory/main.tf line 96, in resource "google_project" "main":
  96: resource "google_project" "main" {

louiskimlevu:
From the console I have a quota error. I sent a quota request but it might be useful to add this info as prerequisite.
![image](https://user-images.githubusercontent.com/57093546/111445808-b8092680-8746-11eb-9cfe-2063ece05d91.png)

m0ps:
That's a limitation of the trial account.
In my case, they requested a top-up account with $50, and then the quota will be increased.

mosqgft:
Same here. You need to open a request to increase quota and put USD 50.

gjj:
Wow thanks for this! I wish I've seen this earlier - it seems like setting this on a fresh Cloud Identity organisation, I hit the limit at `1-org`. Actually, wouldn't it be better to leave a note in `1-org`'s pre-requisites?

rajesh-nitc:
I paid $10 for additional 10 projects. Sharing in case someone wants to pay less for less.

morgante:
Just to be clear, projects themselves are free.

gjj:
Thanks all for the clarifications! Understood that the topup will be applied to future usage, while unlocking more projects at the same time.

While I'm waiting for the quota increase, I have another question: so my `1-org` Cloud Build deployment pipeline created a project halfway and was unable to link my billing account due to the original quota, as it seems like I'm getting `Error 409: Requested entity already exists, alreadyExists.`

Should I delete that specific project manually or just let Terraform do its job once I get my quota increase?

PR #290, opened 2020-11-17
doc: Add note on billing quota exceeded

vovinacci:
This PR adds note on possible exceeded billing quota (I've experienced it myself) to document this issue before troubleshooting guide will be in place.

Closes #269

PR #497, opened 2021-07-08
chore: added link to FAQ in 1-org

gjj:
In this PR,

- I added a note in the prerequisites of `1-org` to inform users that they should request for sufficient projects quota and linked it to the detailed explanation in `FAQ.md`.

The reason for doing so is because I personally did not read the `FAQ.md` document nor the troubleshooting guide (I wasn't aware of those documents until I searched the error codes I was receiving on Google Search) and just followed the instructions as-is. On a fresh Cloud Identity org, `1-org` is where I hit the projects quota.

Now, I understand there are several issues related to this even with a fix in https://github.com/terraform-google-modules/terraform-example-foundation/pull/290 although I'm not sure why that note is gone and there was no mention of any prerequisites to check if you have enough projects quota in the `0-bootstrap` or `1-org`'s prerequisites.

That PR #290 was in November 2020 and even until today, I've noticed - after crawling through existing GitHub issues - that our users are still facing the same problem:

- https://github.com/terraform-google-modules/terraform-example-foundation/issues/378
- https://github.com/terraform-google-modules/terraform-example-foundation/issues/494

It's a simple note, and I believe it would help a lot, especially for people who may be new to cloud security foundations. Appreciate any feedback. Thanks!

### DIFF
--- a/1-org/README.md
+++ b/1-org/README.md
@@ -62,6 +62,7 @@ The purpose of this step is to set up top-level shared folders, monitoring and n
 2. Cloud Identity / Google Workspace group for security admins.
 3. Membership in the security admins group for the user running Terraform.
 4. Security Command Center notifications require that you choose a Security Command Center tier and create and grant permissions for the Security Command Center service account as outlined in [Setting up Security Command Center](https://cloud.google.com/security-command-center/docs/quickstart-security-command-center)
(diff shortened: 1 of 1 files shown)


bharathkkb (approved):
Thanks for the PR  #290 was moved to a section in [troubleshooting doc](https://github.com/terraform-google-modules/terraform-example-foundation/blob/master/docs/TROUBLESHOOTING.md#billing-quota-exceeded)

PR #500, opened 2021-07-10
Add note regarding terraform version for manual steps

### LINKED ISSUE #379: Add note to README regarding terraform version for manual execution
In the Cloud Build instructions for 3-networks and 4-projects(open PR https://github.com/terraform-google-modules/terraform-example-foundation/pull/374) there are steps that require the manual execution of the `shared` folders.

There should be a note in both README files that indicates that the terraform version used in this manual deployment should be the same one that is used in the cloud build, otherwise the user may face a compatibility error if the versions are different when running Cloud Build.

The terraform version used in the cloud build is defined here:
https://github.com/terraform-google-modules/terraform-example-foundation/blob/9b4830e6e32a3c0b99b3f151ba4826808ceb546a/0-bootstrap/main.tf#L117

vovinacci:
Closes #379 

This PR

- Adds notes in `3-networks` and `4-projects` regarding using same Terraform version for manual steps should be the same one that is used in the Cloud Build, otherwise the user may face a compatibility error if the versions are different.
- Fixes typos.

### DIFF
--- a/0-bootstrap/README.md
+++ b/0-bootstrap/README.md
@@ -66,8 +66,7 @@ installed:
 - [Terraform](https://www.terraform.io/downloads.html) version 0.13.7.
 - An existing project which the user has access to be used by terraform-validator.
 
-**Note:** Make sure that you use the same version of Terraform throughout this
-series. Otherwise, you might experience Terraform state snapshot lock errors.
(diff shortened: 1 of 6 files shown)


bharathkkb (approved):
Thanks for the PR @vovinacci

PR #511, opened 2021-07-23
chore: Dependabot for upstream TF modules

### LINKED ISSUE #508: In 1-org + log_sinks.tf I had an issue with 'call to function "list" failed"
Summary
* trying to find the approved way of overcoming issue with log_sinks.tf (or what I am doing that is 'bad')
* trying to understand the updating approach you have for modules (as updating the module is one solution for me)

I had an error when trying to perform terraform apply in "1-org" and there might be multiple options for solving (or I might be doing something 'wrong' - so apols).

File https://github.com/terraform-google-modules/terraform-example-foundation/blob/master/1-org/envs/shared/log_sinks.tf

I got the error:

│ Error: Error in function call
│
│   on .terraform/modules/log_export_to_biqquery/main.tf line 33, in locals:
│   33:   bigquery_options = var.bigquery_options == null ? [] : var.unique_writer_identity == true ? list(var.bigquery_options) : []
│     ├────────────────
│     │ var.bigquery_options is object with 1 attribute "use_partitioned_tables"
│
│ Call to function "list" failed: the "list" function was deprecated in Terraform v0.12 and is no longer available; use tolist([ ... ]) syntax to write a literal list.

The above is due to the following block in log_sinks.tf:

image

So clearly there is the easy fix in syntax to make it work so I wondered if there was a better version of log_export_to_bigquery and was surprised to see there were so many major updates to the module (5.2 / 6.0 and even 7.0.0) whilst this code is  "~> 5.1.0".

Note: I updated my local copy of log_sinks.tf to use "~> 7.0.0" and it overcame the issue (but I have not seen all the code work to know that everything is correct - yet)

Hence my question on what the update policy is as my understanding is that this code is to be the Google opinionated template for how a company could bootstrap their GCP and so I would have envisaged a better tracking to the newest advocated modules.  I was wishing to use the code as a template - but I have concerns after seeing it lag so far behind the latest - and wonder what critical updates it is missing).

Views? Should I move to using the old version of Terraform (but versions.tf does say ">= 0.13" ) ? Am I doing something stupid (quite possible) ?

bharathkkb:
Trying dependabot to keep repo insync with upstream module changes

fixes #508 

Example: https://github.com/bharathkkb/terraform-example-foundation/pulls

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,14 @@
+# To get started with Dependabot version updates, you'll need to specify which
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
+# https://help.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+
+version: 2
+updates:
+  - package-ecosystem: "terraform"
+    directory: "/"
+    schedule:
+      interval: "daily"
+    commit-message:
+      prefix: "chore(deps): "
+    rebase-strategy: "disabled"
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -8,7 +8,90 @@ updates:
   - package-ecosystem: "terraform"
     directory: "0-bootstrap"
     schedule:
-      interval: "daily"
+      interval: "weekly"
+      day: "tuesday"
+      time: "09:00"
     commit-message:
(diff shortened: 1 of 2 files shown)


rjerrems (approved):
LGTM

PR #514, opened 2021-07-28
feat: replace scc gcloud provisioner with native resource

ericyz:


1-org/envs/shared/scc_notification.tf
```diff
@@ -29,28 +29,14 @@ resource "google_pubsub_subscription" "scc_notification_subscription" {
...
-  destroy_cmd_entrypoint = "gcloud"
-  destroy_cmd_body       = <<-EOF
-  scc notifications delete organizations/${var.org_id}/notificationConfigs/${var.scc_notification_name} \
-  --impersonate-service-account ${var.terraform_service_account} \
-  --project "${module.scc_notifications.project_id}" \
-  --quiet
-  EOF
-  skip_download          = var.skip_gcloud_download
+  streaming_config {
+    filter = "category = \"OPEN_FIREWALL\" AND state = \"ACTIVE\""  #var.scc_notification_filter
```
bharathkkb, line 40: I think this should be
```suggestion
    filter = var.scc_notification_filter
```

bharathkkb (changes requested):
Thanks for the PR @ericyz 
Looks like the lint test is failing.

1-org/envs/shared/scc_notification.tf
daniel-cit: Maybe the current *default* value for the  scc_notification_filter 

https://github.com/terraform-google-modules/terraform-example-foundation/blob/a9145f48fbf53df3fa9aaabc31b887b6e1a9f11d/1-org/envs/shared/variables.tf#L81-L86

is over escaped and the `\\\"` could be shortened to just `\"`

ericyz: Oops, I forgot to replace the testing value and will update it in next commit

ericyz:
A related upstream issue on this change

https://github.com/hashicorp/terraform-provider-google/issues/9670

bharathkkb (approved):
Thanks @ericyz

PR #553, opened 2021-09-02
chore: issue 487. clarify "named" branches vs. others

mark1000:


mark1000:
Am I correct in understanding that only the lint test matters for this change?

1-org/README.md
```diff
@@ -166,12 +166,15 @@ Run `terraform output cloudbuild_project_id` in the `0-bootstrap` folder to see
    git add .
    git commit -m 'Your message'
    ```
-1. Push your plan branch to trigger a plan. For this command, the branch `plan` is not a special one. Any branch which name is different from `development`, `non-production` or `production` will trigger a Terraform plan.
+1. Push your plan branch to trigger a plan for all environments. Because the
+   _plan_ branch is not a [named branch](./docs/FAQ.md), pushing your _plan_
```
morgante, line 170: We should refer to this as a "named environment branch" so there's at least some contextual info on why it exists.

morgante (changes requested):

1-org/README.md
mark1000: Like that?

bharathkkb:
@mark1000 can we switch to `named environment branch` everywhere to prevent confusion btw `named branch`  and `named environment branch` ?

PR #559, opened 2021-09-14
upgrading cloud router module version

williamtsoi1:
this is needed to retain compatibility with Terraform version >0.14


rjerrems (approved):
LGTM - Thanks @williamtsoi1 

cc @bharathkkb

PR #582, opened 2021-11-04
Fixes #486

### LINKED ISSUE #486: [UX] Pushing to empty repos created in CSR with init.defaultBranch set
Following [1-org instructions step 5](https://github.com/terraform-google-modules/terraform-example-foundation/tree/master/1-org#deploying-with-cloud-build), after cloning an empty CSR repo and making the first commit push `git push --set-upstream origin master` may return `error: src refspec master does not match any`. This could be due to `init.defaultBranch` being set to something other than master (`main` for instance)

We should update docs to make users aware of this and switch our defaults to explicitly be main `git push --set-upstream origin main`. I don't anticipate any other code changes as we name our real main branches same as envs.

mark1000:


bharathkkb (commented):
@mark1000 it looks like there are no changes in this diff
https://github.com/terraform-google-modules/terraform-example-foundation/pull/582/files

PR #583, opened 2021-11-04
Fixes #487

### LINKED ISSUE #487: [UX] Indicate what TF command will run for push to a given branch
Alongside [existing steps](https://github.com/terraform-google-modules/terraform-example-foundation/tree/master/2-environments#:~:text=push%20your%20plan%20branch%20to%20trigger%20a%20plan%20for%20all%20environments) for git push to a branch, we should also document what Terraform command will be run by the CI. This will

- cement the idea of what commands are applied to main vs non main branches 
- allow users who are unfamiliar with TF to understand if it failed/passed during plan or apply

mark1000:


bharathkkb:
@mark1000 it looks like there are no changes in this diff
https://github.com/terraform-google-modules/terraform-example-foundation/pull/583/files

PR #588, opened 2021-11-11
Add Assured workload folder for production environment

daniel-cit:
This PR  creates a folder,`fdr-production-assured-workload`, under the `fdr-production` folder  and configure  Assured Workload for this new folder

### DIFF
--- a/2-environments/envs/production/assured_workload.tf
+++ b/2-environments/envs/production/assured_workload.tf
@@ -14,13 +14,6 @@
  * limitations under the License.
  */
 
-resource "google_folder" "assured_workload" {
-  count = var.assured_workload_configuration != null && var.assured_workload_configuration.enabled ? 1 : 0
-
-  display_name = "${var.folder_prefix}-assured-workload"
-  parent       = module.env.env_folder
-}
-
 resource "google_assured_workloads_workload" "production" {
   provider = google-beta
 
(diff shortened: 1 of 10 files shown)


2-environments/envs/production/assured_workload.tf
```diff
@@ -0,0 +1,33 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+locals {
+  assured_workload_display_name = "${var.folder_prefix}-assured-workload"
+  enable_assured_workload       = var.assured_workload_configuration != null && var.assured_workload_configuration.enabled
```
bharathkkb, line 19: Cant we do?
```suggestion
  enable_assured_workload       = lookup(var.assured_workload_configuration, "enabled", false)
```

2-environments/envs/production/outputs.tf
```diff
@@ -38,3 +38,8 @@ output "env_secrets_project_id" {
   description = "Project for environment related secrets."
   value       = module.env.env_secrets_project_id
 }
+
+output "assured_workload_folder_display_name" {
+  description = "Assured Workload folder display name."
+  value       = local.enable_assured_workload ? local.assured_workload_display_name : ""
```
bharathkkb, line 44: `id` maybe more useful for end users

morgante: Do we need an output at all?

bharathkkb: @morgante I was thinking of usage in tests (using [gcloud assured workloads describe](https://cloud.google.com/sdk/gcloud/reference/assured/workloads/describe)) as the resource seems to have a generated UUID as name.

morgante: Ah that makes sense. I agree we should return ID then.

2-environments/envs/production/assured_workload.tf
daniel-cit: fixed

2-environments/envs/production/outputs.tf
daniel-cit: the ID of the workload added an an output + the folder id that contains the workload

PR #618, opened 2022-01-05
fix infra-pipeline gcs location (4-projects)

plassen:
Hi again, it seems the supporting buckets for the infra pipeline, defined on 4-projects, are always created on the default location. I've added the location_gcs variable as input parameter to allow customization. Thanks!


bharathkkb (changes requested):
Thanks for the PR @plassen 
We should [define the variable](https://github.com/terraform-google-modules/terraform-example-foundation/blob/1fad10b2af93069a7369cd100a3fcf2cddaa6f74/4-projects/business_unit_1/shared/variables.tf) here as well

Issue #608, opened 2021-12-22
3-networks - terraform-validator constraint violation : "Flow logs are disabled in subnetwork"


plassen:
### TL;DR

(Disclaimer: as I don't see any issue or note about this error around here, so probably the problem is mine)

When applying the Cloud Build 'plan' trigger to the gcp-networks repo, I get validation errors for the subnet creation objects, complaining that the flow logging option is disabled, when in fact, the 'log_config' block is defined on the generated tfplan.

The manual deployment of the 'shared' section went well (step 11) because it doesn't mention applying terraform-validator to the generated tfplan. When doing so, it also complains about it :)

### Expected behavior

should provide no violations because var.subnetworks_enable_logging is with its default value 'true'

### Observed behavior

[log-a2091d83-0a40-4a3c-b444-acd7fcb392c1.txt](https://github.com/terraform-google-modules/terraform-example-foundation/files/7764188/log-a2091d83-0a40-4a3c-b444-acd7fcb392c1.txt)

### Terraform Configuration

```hcl
details from the files on the 'env/development' subfolder:

variables.tf:
variable "subnetworks_enable_logging" {
  type        = bool
  description = "Toggle subnetworks flow logging for VPC Subnetworks."
  default     = true
}

sample from main.tf:
module "restricted_shared_vpc" {
  source                           = "../../modules/restricted_shared_vpc"
  project_id                       = local.restricted_project_id
  project_number                   = local.restricted_project_number
  environment_code                 = local.environment_code
  access_context_manager_policy_id = var.access_context_manager_policy_id
  restricted_services              = ["bigquery.googleapis.com", "storage.googleapis.com"]
  members                          = ["serviceAccount:${var.terraform_service_account}"]
  private_service_cidr             = local.restricted_private_service_cidr
  org_id                           = var.org_id
  parent_folder                    = var.parent_folder
  bgp_asn_subnet                   = local.bgp_asn_number
  default_region1                  = var.default_region1
  default_region2                  = var.default_region2
  domain                           = var.domain
  windows_activation_enabled       = var.windows_activation_enabled
  dns_enable_inbound_forwarding    = var.dns_enable_inbound_forwarding
  dns_enable_logging               = var.dns_enable_logging
  firewall_enable_logging          = var.firewall_enable_logging
  optional_fw_rules_enabled        = var.optional_fw_rules_enabled
  nat_enabled                      = var.nat_enabled
[...]

PR #622, opened 2022-01-21
Updates gcp_network_enable_flow_logs_v1.yaml template.

amandakarina:
Fix issue #608 

Updates subnetwork flow logs template to new log configuration format.


bharathkkb:
Thanks Amanda, can we also open a PR to fix this in the policy-repo

amandakarina:
This fix should be done in the [policy-repo](https://github.com/GoogleCloudPlatform/policy-library/issues/414) and be compatible with old and new version of enabling flow logs.

After this, foundation should be updated.

Issue #628, opened 2022-02-05
1-org Cloud Build fails with googleapi: Error 403: The caller does not have permission, forbidden


ipv1337:
### TL;DR

Similar to issue #620, but mine errors while trying to fetch retrieving IAM policy. I've solved this with the missing IAM for the cloud build SA, but I wanted to open an issue to attach my PR to.

### Expected behavior

Following the TFV example for setting ip 1-org should work in Cloud Build as described in the docs. https://github.com/terraform-google-modules/terraform-example-foundation/tree/master/1-org#deploying-with-cloud-build

### Observed behavior

After pushing my plan branch, cloud build errors with the following:

```
Step #1 - "tf plan validate all": Error: converting tfplan to CAI assets: adding resource changes to converter: adding resource create or update fetching asset: Error retrieving IAM policy for organization "XXXXXXXXX": googleapi: Error 403: The caller does not have permission, forbidden
Step #1 - "tf plan validate all": Usage:
Step #1 - "tf plan validate all":   terraform-validator validate <tfplan> [flags]
Step #1 - "tf plan validate all": 
Step #1 - "tf plan validate all": Flags:
Step #1 - "tf plan validate all":       --ancestry string      Override the ancestry location of the project when validating resources
Step #1 - "tf plan validate all":   -h, --help                 help for validate
Step #1 - "tf plan validate all":       --offline              Do not make network requests
Step #1 - "tf plan validate all":       --output-json          Print violations as JSON
Step #1 - "tf plan validate all":       --policy-path string   Path to directory containing validation policies
Step #1 - "tf plan validate all":       --project string       Provider project override (override the default project configuration assigned to the google terraform provider when validating resources)
Step #1 - "tf plan validate all": 
Step #1 - "tf plan validate all": Global Flags:
Step #1 - "tf plan validate all":       --verbose   Log output to stderr
```

### Terraform Configuration

```hcl
Latest code from https://github.com/terraform-google-modules/terraform-example-foundation/tree/master/1-org
```

### Terraform Version

```sh
james@cloudshell:~/.../envs/shared$ terraform -version
Terraform v0.13.7
```

### Additional information

Fix 0-bootstrap/main.tf and added this block:

```
resource "google_organization_iam_member" "org_cb_sa_iam_viewer" {
  count  = var.parent_folder == "" ? 1 : 0
  org_id = var.org_id
  role   = "roles/iam.organizationRoleViewer"
[...]

Issue #546, opened 2021-08-19
Update permissions required for TFV


bharathkkb:
TFV has been using the CB default SA (not Terraform SA) which did not have any additional permissions, however now `v2021-04-07+` it looks like it needs Security Reviewer at the org level to pull IAM policy otherwise this error is thrown for `1-org` 

```
 Error: converting tfplan to CAI assets: adding resource changes to converter: adding resource create/update/no-op fetching asset: Error retrieving IAM policy for organization "ORG_ID": googleapi: Error 403: The caller does not have permission, forbidden
```

We should add Security Reviewer to the CB SA once confirmed via https://github.com/GoogleCloudPlatform/terraform-validator/issues/290

Longer term (v3) we maybe looking at configuring custom SAs via [CB trigger](https://cloud.google.com/build/docs/securing-builds/configure-user-specified-service-accounts#running_builds_using_build_triggers)

pharveybattelle:
As a workaround manually add Security Reviewer to the CB SA (CICD_PROJECT_NUM@cloudbuild.gserviceaccount.com) at the organization level.
I have confirmed this work around works in our environment.

bharathkkb:
Short term we can use `GOOGLE_IMPERSONATE_SERVICE_ACCOUNT` rather than granting CB SA permissions. Long term we will be switching to custom SA from impersonation approach.

bharathkkb:
Since we have a workaround in place, @amandakarina will explore using user-specified service account cloud build SA rather than SA impersonation which will help us get rid of provider impersonation and solve the TFV auth issue.

bharathkkb:
Closing this since we now have user defined SA in pipeline and GOOGLE_IMPERSONATE_SERVICE_ACCOUNT is used for local which is respected by gcloud vet. @daniel-cit lmk if this is still an issue.

PR #629, opened 2022-02-05
fix: Issue #628 Failed Cloud Build plan in 1-org using cloud build SA from TFV

ipv1337:
This fixed my issue #628.

### DIFF
--- a/0-bootstrap/main.tf
+++ b/0-bootstrap/main.tf
@@ -183,6 +183,13 @@ resource "google_organization_iam_member" "org_cb_sa_iam_viewer" {
   member = "serviceAccount:${data.google_project.cloudbuild.number}@cloudbuild.gserviceaccount.com"
 }
 
+resource "google_folder_iam_member" "org_cb_sa_iam_viewer" {
+  count  = var.parent_folder != "" ? 1 : 0
+  org_id = var.org_id
+  role   = "roles/iam.organizationRoleViewer"
+  member = "serviceAccount:${data.google_project.cloudbuild.number}@cloudbuild.gserviceaccount.com"
+}
+
 resource "google_organization_iam_member" "org_cb_sa_browser" {
   count  = var.parent_folder == "" ? 1 : 0
   org_id = var.org_id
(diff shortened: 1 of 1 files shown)


bharathkkb (changes requested):
Thanks for the PR, could you also add a google_folder_iam_member binding if `var.parent_folder != ""`?

ipv1337:
> Thanks for the PR, could you also add a google_folder_iam_member binding if `var.parent_folder != ""`?

Let me know if the recent commit is what you are looking for @bharathkkb.

0-bootstrap/main.tf
```diff
@@ -176,6 +176,20 @@ data "google_project" "cloudbuild" {
   depends_on = [module.cloudbuild_bootstrap.csr_repos]
 }
 
+resource "google_organization_iam_member" "org_cb_sa_iam_viewer" {
+  count  = var.parent_folder == "" ? 1 : 0
+  org_id = var.org_id
+  role   = "roles/iam.organizationRoleViewer"
```
bharathkkb, line 182:
```suggestion
  role   = "roles/iam.securityReviewer"
```

0-bootstrap/main.tf
```diff
@@ -176,6 +176,20 @@ data "google_project" "cloudbuild" {
...
+  member = "serviceAccount:${data.google_project.cloudbuild.number}@cloudbuild.gserviceaccount.com"
+}
+
+resource "google_folder_iam_member" "org_cb_sa_iam_viewer" {
+  count  = var.parent_folder != "" ? 1 : 0
+  folder = var.parent_folder
+  role   = "roles/iam.organizationRoleViewer"
```
bharathkkb, line 189:
```suggestion
  role   = "roles/iam.securityReviewer"
```

bharathkkb (commented):
It looks like we will actually need `roles/iam.securityReviewer` as [TFV will need to get IAM policy associated with any resource](https://github.com/GoogleCloudPlatform/terraform-validator/blob/8f1cc1a0d8d7feba66d8eb67f16467a3d930ab82/docs/user_guide.md#required-permissions) x-ref: https://github.com/terraform-google-modules/terraform-example-foundation/issues/546. It also looks like `organizationRoleViewer` can only be granted at the org level

bharathkkb (approved):

PR #630, opened 2022-02-07
fix: data_access_logs_enabled now enables read and write audit logs, defaults to false for cost savings

amandakarina:
Hey folks, 

This pull request comment the code where enable all the audit logs, due to the costs can be generated.

I also added a disclaimer of charges and to remove the comment of the code to enable all of the audit logs.

### DIFF
--- a/1-org/envs/shared/iam.tf
+++ b/1-org/envs/shared/iam.tf
@@ -23,12 +23,19 @@ resource "google_organization_iam_audit_config" "org_config" {
   org_id  = var.org_id
   service = "allServices"
 
-  audit_log_config {
-    log_type = "DATA_READ"
-  }
-  audit_log_config {
-    log_type = "DATA_WRITE"
-  }
+  ###################################################################################################
+  ### Audit logs can generate costs, to know more about it,
+  ### check the official documentation: https://cloud.google.com/stackdriver/pricing#logging-costs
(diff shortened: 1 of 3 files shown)


bharathkkb (commented):
Thanks for the PR @amandakarina 
I think the usage of `data_access_logs_enabled` maybe confusing with this change. Is `data_access_logs_enabled` not sufficient?

1-org/envs/shared/iam.tf
```diff
@@ -18,32 +18,50 @@
...
-    log_type = "DATA_WRITE"
+  ###################################################################################################
+  ### Audit logs can generate costs, to know more about it,
+  ### check the official documentation: https://cloud.google.com/stackdriver/pricing#logging-costs
+  ### To know more about audit logs, you can find more infos
+  ### here https://cloud.google.com/logging/docs/audit/configure-data-access
+  ### To enable DATA_READ and DATA_WRITE audit logs, set `data_access_logs_enabled` to true
+  ####################################################################################################
+  dynamic "audit_log_config" {
+    for_each = local.enabling_data_logs
```
bharathkkb, line 38: nit: this would let us use a single dynamic block
```suggestion
    for_each = setunion(local.enabling_data_logs, ["ADMIN_READ"])
```

bharathkkb (approved):

PR #667, opened 2022-03-28
set the location for cloud build related buckets in step4 based in the default region

### LINKED ISSUE #666: Buckets for the infra pipeline create in step 4 should used the default region provided by the user
### TL;DR

The Buckets created for for the infra pipeline create in step 4 :

- Terraform state bucket
- Cloud build artifacts bucket
- Cloud build default bucket

should used the default region provided by the user

### Expected behavior

The three buckets
- Terraform state bucket
- Cloud build artifacts bucket
- Cloud build default bucket

Should be create in the `default_region` provided by the user

### Observed behavior

Buckets
- Terraform state bucket
- Cloud build artifacts bucket

are always created in the region  `us-central1`

And the Cloud build default bucket (source staging bucket) is create in the multi-region `us`

### Terraform Configuration

```hcl
repo was deployed using the defaults values but changing the regions so that they are in Europe.
```

### Terraform Version

```sh
Terraform v1.0.6
on linux_amd64
```

### Additional information

The  Cloud build default bucket is created automatically in the `us` location when the cloud build API is used by the first time.
A bucket with the name `<PROJECT_ID>_cloudbuild` can be created in the selected region before the creation of the first build.
This bucket will then be used by cloud build.

daniel-cit:
Fixes #666

### DIFF
--- a/4-projects/modules/infra_pipelines/main.tf
+++ b/4-projects/modules/infra_pipelines/main.tf
@@ -15,12 +15,14 @@
  */
 
 locals {
-  gar_repo_name        = var.gar_repo_name != "" ? var.gar_repo_name : format("%s-%s", var.project_prefix, "tf-runners")
-  gar_name             = split("/", google_artifact_registry_repository.tf-image-repo.name)[length(split("/", google_artifact_registry_repository.tf-image-repo.name)) - 1]
-  created_csrs         = toset([for repo in google_sourcerepo_repository.app_infra_repo : repo.name])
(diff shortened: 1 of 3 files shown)


4-projects/modules/infra_pipelines/main.tf
```diff
@@ -66,13 +68,30 @@ resource "google_storage_bucket" "cloudbuild_artifacts" {
   }
 }
 
+resource "google_storage_bucket" "cloudbuild_bucket" {
```
bharathkkb, line 71: This seems like we could refactor to use for_each over `merge(local.artifact_buckets, local.state_buckets, local.cloudbuild_bucket)`?

bharathkkb (approved):
minor nit, LGTM otherwise

bharathkkb (approved):

PR #715, opened 2022-05-03
fix: add propagation time to test/setup

apeabody:


apeabody:
I believe the CI should be unblocked #717

test/setup/main.tf
```diff
@@ -49,3 +49,10 @@ module "project" {
     "billingbudgets.googleapis.com",
   ]
 }
+
+# propagation time
+resource "time_sleep" "wait_180_seconds" {
```
daniel-cit, line 54: the integration build has a step to wait for propagation (services, iam) from the setup

https://github.com/terraform-google-modules/terraform-example-foundation/blob/3e2ca413d2fd57c309748109f56008be2503ff9a/build/int.cloudbuild.yaml#L25-L27

maybe the sleep time in the step in `int.cloudbuild.yaml` could be changed if it is not enough anymore

daniel-cit: looking at some of the build errors with `SERVICE_DISABLED`, it looks like that the APIs not enabled errors are in the `seed` and `cicd` projects that are created by the bootstrap module. 
Maybe this change could be implemented there or in the project-factory module

daniel-cit: looking at the error messages, the project that is been reported as not having the APIs enable is **the project created in the setup step**. 
Adding the extra minutes to the [int.cloudbuild.yaml](https://github.com/terraform-google-modules/terraform-example-foundation/blob/3e2ca413d2fd57c309748109f56008be2503ff9a/build/int.cloudbuild.yaml#L25-L27) file should reduce the problem

apeabody: Hi @daniel-cit thanks for the feedback and suggestions!  I'll review it sometime next week.  Cheers!

apeabody: Thanks @daniel-cit - I see https://github.com/terraform-google-modules/terraform-example-foundation/blob/90f15f28c03f391dc95aaa843f1a5727772bd8e4/build/int.cloudbuild.yaml#L27 was committed last week, so I'll close out this PR.

apeabody:
Close in favor of https://github.com/terraform-google-modules/terraform-example-foundation/blob/90f15f28c03f391dc95aaa843f1a5727772bd8e4/build/int.cloudbuild.yaml#L27

PR #762, opened 2022-07-28
chore: build stability fixes

daniel-cit:
build stability:
- Add random `project_prefix` in test setup output to prevent project name collision
- Retry in common recoverable errors
- Fix google terraform provider `request_timeout` to be a valid duration ( `120s` => `2m`)
- Add  API warm up in bootstrap step
- Add  API warm up in org step


test/setup/main.tf
```diff
@@ -27,7 +32,7 @@ module "project" {
   source  = "terraform-google-modules/project-factory/google"
   version = "~> 13.0"
 
-  name              = "ci-foundation"
+  name              = "ci-foundation-${local.project_prefix}"
```
apeabody, line 35: This would be a suffix :)

daniel-cit: good point

daniel-cit: fixed

4-projects/modules/infra_pipelines/versions.tf
```diff
@@ -19,11 +19,11 @@ terraform {
   required_providers {
     google = {
       source  = "hashicorp/google"
-      version = ">= 3.50"
+      version = "~> 4.30.0"
```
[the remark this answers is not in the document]
daniel-cit: @apeabody this is a temporary fix until a fix for a bug in 4.31.0 is released

test/setup/outputs.tf
```diff
@@ -68,6 +68,10 @@ output "monitoring_workspace_users" {
   value = var.group_email
 }
 
+output "project_prefix" {
```
apeabody, line 71:
```suggestion
output "project_suffix" {
```

test/setup/main.tf
```diff
@@ -14,20 +14,25 @@
...
+  project_randomness = "${replace(substr(local.random_id, 0, 1), "/[0123456789]/", "z")}${substr(local.random_id, 1, 3)}"
+}
+
+resource "random_id" "source" {
   byte_length = 2
 }
 
 resource "google_folder" "test_folder" {
-  display_name = "test_foundation_folder_${random_id.random_project_id_suffix.hex}"
+  display_name = "test_foundation_folder_${random_id.source.hex}"
```
apeabody, line 27: I would avoid using hex if possible as that limits you to 0-9a-f rather than 0-9a-z.  Consider using `random_string` as exampled above.

test/setup/outputs.tf
daniel-cit: this output need to be `project_prefix` because it is replacing this env var

https://github.com/terraform-google-modules/terraform-example-foundation/blob/d1a56d408cdb6ef95da51c685d6beaf5f6a0b07d/build/int.cloudbuild.yaml#L156

that is used to define which project prefix will be used in all steps

test/setup/main.tf
```diff
@@ -14,20 +14,25 @@
  * limitations under the License.
  */
 
-resource "random_id" "random_project_id_suffix" {
+locals {
+  random_id          = replace(lower(random_id.source.id), "/[_-]/", "t")
+  project_randomness = "${replace(substr(local.random_id, 0, 1), "/[0123456789]/", "z")}${substr(local.random_id, 1, 3)}"
+}
+
+resource "random_id" "source" {
```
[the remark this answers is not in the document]
daniel-cit: @apeabody  Great idea :)

test/integration/bootstrap/bootstrap_test.go
```diff
@@ -77,6 +107,32 @@ func TestBootstrap(t *testing.T) {
...
+				"cloudbuild.googleapis.com",
+				"sourcerepo.googleapis.com",
+				"cloudkms.googleapis.com",
+				"bigquery.googleapis.com",
+				"accesscontextmanager.googleapis.com",
+				"securitycenter.googleapis.com",
+				"servicenetworking.googleapis.com",
+				"billingbudgets.googleapis.com",
+			} {
+				retry.DoWithRetry(t, fmt.Sprintf("checking if %s API is enabled in project %s", api, projectID), 5, 2*time.Minute, func() (string, error) {
```
bharathkkb, line 128: nit: lets use https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/blob/339cb248327248caca129bb31e309dc8afcae345/infra/blueprint-test/pkg/utils/poll.go#L27

test/integration/bootstrap/bootstrap_test.go
```diff
@@ -34,6 +40,30 @@ func getResultFieldStrSlice(rs []gjson.Result, field string) []string {
 	return s
 }
 
+func checkAPIEnabled(t *testing.T, projectID, api string) (string, error) {
```
bharathkkb, line 43: Any reason why we need the string?
```suggestion
func checkAPIEnabled(t *testing.T, projectID, api string) (error) {
```

test/integration/org/org_test.go
```diff
@@ -44,6 +49,30 @@ func getResultFieldStrSlice(rs []gjson.Result, field string) []string {
 	return s
 }
 
+func checkAPIEnabled(t *testing.T, projectID, api string) (string, error) {
```
bharathkkb, line 52: Rather than duplicating code, we could centralize this in a testutils package and reference it here.

test/integration/bootstrap/bootstrap_test.go
daniel-cit: replace with `utils.Poll`

test/integration/bootstrap/bootstrap_test.go
daniel-cit: the return `(string, error)` was part of the signature of `retry.DoWithRetry`.
changed to `(bool, error)` because of the signature of `utils.Poll`

test/integration/org/org_test.go
daniel-cit: moved to `testutils` package

test/integration/testutils/strings.go
```diff
@@ -0,0 +1,24 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package testutils
+
+import (
+	"strings"
+)
+
+func GetLastSplitElement(value string, sep string) string {
```
bharathkkb, line 21: These are pretty useful, we should upstream them :)

bharathkkb (approved):

PR #765, opened 2022-08-03
Feature/validate requirements script

mauro-cit:
This PR is adding a script that will make sure to validate wheter the environment that will run the terraform foundation steps are configured with the expected requirements. These requirements are:

- Checking if the machine have installed: Git, Terraform and gcloud sdk
- Check if the version of git and gcloud sdk are above a minimum expected
- Check if Terraform version is exact the one expected
- Check if the user have runned `gcloud auth login` and `gcloud auth application-default login` locally with a valid account
- Check if the end user credential that will be used to run the terraform scripts contains the required Organization and Billing account level roles
- Check if the bootstrap terraform.tfvars is found and have the required values correctly filled

### DIFF
--- a/scripts/validate-requirements.sh
+++ b/scripts/validate-requirements.sh
@@ -1,0 +1,69 @@
+#!/bin/bash
+# -------------------------- Variables --------------------------
+# Versions of the installators
+# TODO
+# Must obtain the correct required versions
+TF_VERSION="v1.2.1"
+GCLOUD_SDK_VERSION="380.0.0"
+GIT_VERSION="2.25.1"
+# -------------------------- Funcions ---------------------------
+# TODO
+# Must update this comparission mechanism to compare if version is HIGHER not only EQUAL
+# ref: https://stackoverflow.com/questions/4023830/how-to-compare-two-strings-in-dot-separated-version-format-in-bash
+# Function to validate the Terraform installation and version
+function validate_terraform(){
+    which terraform
+    if [ $? -ne 0 ]; then
+        echo "Terraform not found."
+        echo "Visit https://learn.hashicorp.com/tutorials/terraform/install-cli and follow the instructions to install Terraform."
+        exit 1
+    else
+        terraform version | grep -e $TF_VERSION
+        if [ $? -ne 0 ]; then
+            echo "An error was found with the Terraform installation."
+            echo "Version required is $TF_VERSION"
+            echo "Visit https://learn.hashicorp.com/tutorials/terraform/install-cli and follow the instructions to install Terraform."
+            exit 1
+        fi
+    fi
+}
+# Function to validate the Google Cloud SDK installation and version
+function validate_gcloud(){
+    which gcloud
+    if [ $? -ne 0 ]; then
+        echo "Gcloud not found."
+        echo "Visit https://cloud.google.com/sdk/docs/install and follow the instructions to install gcloud CLI."
+        exit 1
+    else
+        gcloud version | grep -e $GCLOUD_SDK_VERSION
+        if [ $? -ne 0 ]; then
+            echo "An error was found with the Google Cloud SDK installation."
+            echo "Version required is $GCLOUD_SDK_VERSION"
+            echo "Visit https://cloud.google.com/sdk/docs/install and follow the instructions to install gcloud CLI."
+            exit 1
+        fi
+    fi
+}
+# Function to validate the Git installation and version
+function validate_git(){
+    which git
+    if [ $? -ne 0 ]; then
+        echo "git not found."
+        echo "Visit https://git-scm.com/book/en/v2/Getting-Started-Installing-Git and follow the instructions to install Git."
+        exit 1
+    else
+        git version | grep -e $GIT_VERSION
+        if [ $? -ne 0 ]; then
+            echo "An error was found with the git installation."
+            echo "Version required is $GIT_VERSION"
+            echo "Visit https://git-scm.com/book/en/v2/Getting-Started-Installing-Git and follow the instructions to install Git."
+            exit 1
+        fi
+    fi
+}
+echo "Validating Terraform installation..."
+validate_terraform
+echo "Validating Google Cloud SDK installation..."
+validate_gcloud
+echo "Validating GIT installation..."
+validate_git
--- a/scripts/validate-requirements.sh
+++ b/scripts/validate-requirements.sh
@@ -112,6 +112,16 @@ function validate_gcloud_configuration(){
         exit 1
     fi
     rm -Rf end-user-credential-output.txt
+
+    gcloud auth application-default print-access-token 2> application-default-credential-output.txt
+    cat application-default-credential-output.txt | grep "Could not automatically determine credentials"
+    if [ $? -eq 0 ]; then
+        echo "You must configure an Application Default Credential."
+        echo "Visit https://cloud.google.com/sdk/gcloud/reference/auth/application-default/login and follow the instructions to authorize gcloud to access the Cloud Platform with Google user credentials."
+        rm -Rf application-default-credential-output.txt
+        exit 1
+    fi
+    rm -Rf application-default-credential-output.txt
 }
 
 # Function to validate the Configuration of the Gcloud CLI
--- a/scripts/validate-requirements.sh
+++ b/scripts/validate-requirements.sh
@@ -23,6 +23,11 @@ TF_VERSION="0.13.7"
 GCLOUD_SDK_VERSION="319.0.0"
 GIT_VERSION="2.25.1"
 
+# Input variables
+END_USER_CREDENTIAL=""
+ORGANIZATION_ID=""
+BILLING_ACCOUNT=""
+
 # Collect the errors
 ERRORS=""
 # -------------------------- Funcions ---------------------------
@@ -193,10 +198,6 @@ function validate_bootstrap_step(){
 
 function main(){
 
-    END_USER_CREDENTIAL="$u"
-    ORGANIZATION_ID="$o"
-    BILLING_ACCOUNT="$b"
-
     echo "Validating Terraform installation..."
     validate_terraform
 
@@ -237,18 +238,17 @@ usage() {
     exit 1
 }
 
-
 # Check for input variables
 while getopts ":o:b:u:" OPT; do
   case ${OPT} in
     o )
-      o=$OPTARG
+      ORGANIZATION_ID=$OPTARG
       ;;
     b )
-      b=$OPTARG
+      BILLING_ACCOUNT=$OPTARG
       ;;
     u )
-      u=$OPTARG
+      END_USER_CREDENTIAL=$OPTARG
       ;;
     : )
       echo
@@ -265,7 +265,7 @@ done
 shift $((OPTIND -1))
 
 # Check for required input variables
-if [ -z "${o}" ] || [ -z "${b}" ]|| [ -z "${u}" ]; then
+if [ -z "${ORGANIZATION_ID}" ] || [ -z "${BILLING_ACCOUNT}" ]|| [ -z "${END_USER_CREDENTIAL}" ]; then
   echo
   echo " Error: -o <organization id>, -b <billing project> and -u <end user email> required."
   usage
--- a/scripts/validate-requirements.sh
+++ b/scripts/validate-requirements.sh
@@ -160,7 +160,7 @@ function check_billing_account_roles(){
         --format="table(bindings.role)" > billing-account-roles.txt
 
     lines=$(cat billing-account-roles.txt | grep -e roles/billing.admin | wc -l)
-    if [ $lines -ne 1 ]; then
+    if [ "$lines" -ne 1 ]; then
         echo "The User must have the Billing Account Role billing.admin"
         ERRORS+=$'There are missing billing account level roles on the Credential.\n'
     fi
@@ -173,7 +173,7 @@ function validate_bootstrap_step(){
     if [ ! -f "$FILE" ]; then
         echo "$FILE does not exists."
     else
-        if [ $(grep -c "REPLACE_ME" $FILE) != 0 ]; then
+        if [ '$(grep -c "REPLACE_ME" $FILE)' != 0 ]; then
             echo "$FILE must have required values fullfiled."
             ERRORS+=$'terraform.tfvars file must be correctly fullfiled for 0-bootstrap step.\n'
         fi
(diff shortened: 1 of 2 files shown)


scripts/validate-requirements.sh
```diff
@@ -0,0 +1,274 @@
...
+
+# Validate the Terraform installation and version
+function validate_terraform(){
+
+    if [ ! "$(command -v terraform )" ]; then
+        echo "Terraform not found."
+        echo "Visit https://learn.hashicorp.com/tutorials/terraform/install-cli and follow the instructions to install Terraform."
+        ERRORS+=$'Terraform not found\n'
+    else
+        TERRAFORM_CURRENT_VERSION=$(terraform version -json | jq -r .terraform_version)
```
bharathkkb, line 78: Let validate jq is also installed as one of the initial steps.

scripts/validate-requirements.sh
```diff
@@ -0,0 +1,274 @@
...
+# Validate the Google Cloud SDK installation and version
+function validate_gcloud(){
+    if [ ! "$(command -v gcloud)" ]; then
+        echo "gcloud CLI not found."
+        echo "Visit https://cloud.google.com/sdk/docs/install and follow the instructions to install gcloud CLI."
+        ERRORS+=$'gcloud not found.\n'
+    else
+        GCLOUD_CURRENT_VERSION=$(gcloud version --format=json | jq -r '."Google Cloud SDK"')
+        if [ "$(compare_version "$GCLOUD_CURRENT_VERSION" "$GCLOUD_SDK_VERSION")" -eq 2 ]; then
+            echo "An incompatibly was found in gcloud version."
```
bharathkkb, line 97: nit:
```suggestion
            echo "An incompatible gcloud version was found."
```

bharathkkb: A few places where this can be fixed below. I might even move this to a func that takes in install URL and minimum version and echoes this warning.

scripts/validate-requirements.sh
```diff
@@ -0,0 +1,274 @@
...
+GIT_VERSION="2.25.1"
+
+# Input variables
+END_USER_CREDENTIAL=""
+ORGANIZATION_ID=""
+BILLING_ACCOUNT=""
+
+# Collect the errors
+ERRORS=""
+# -------------------------- Funcions ---------------------------
```
bharathkkb, line 33:
```suggestion
# -------------------------- Functions ---------------------------
```

scripts/validate-requirements.sh
```diff
@@ -0,0 +1,274 @@
...
+
+# Compare two semantic versions
+function compare_version(){
```
bharathkkb, line 36: What are expected returns?

scripts/validate-requirements.sh
```diff
@@ -0,0 +1,274 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Usage:
+# bash scripts/validate-requirements.sh "END_USER_EMAIL" "ORGANIZATION_ID" "BILLING_ACCOUNT_ID"
+
+# -------------------------- Variables --------------------------
+# Expected versions of the installers
```
bharathkkb, line 21: Are we requiring these as minversions?

scripts/validate-requirements.sh
```diff
@@ -0,0 +1,274 @@
...
+    if [[ "$1" == "$2" ]]; then
+        echo 0
+        return 0
+    fi
+
+    local IFS=.
+    local i
+    local version1=("$1")
+    local version2=("$2")
+    for ((i=${#version1[@]}; i<${#version2[@]}; i++))
```
bharathkkb, line 46: Comments for this logic will make it easier to follow

scripts/validate-requirements.sh
```diff
@@ -0,0 +1,274 @@
...
+        fi
+    fi
+}
+
+# Validate the Git installation and version
+function validate_git(){
+    if [ ! "$(command -v git)" ]; then
+        echo "git not found."
+        echo "Visit https://git-scm.com/book/en/v2/Getting-Started-Installing-Git and follow the instructions to install Git."
+        ERRORS+=$'git not found.\n'
```
bharathkkb, line 110: Can we exit early in these cases? We prefer early exit as opposed to wrapping in else making it less readable.

scripts/validate-requirements.sh
```diff
@@ -0,0 +1,274 @@
...
+    fi
+
+    if ! git config init.defaultBranch | grep "main" >/dev/null ; then
+        echo "git defaul branch must be configured."
+        echo "See the instructions at https://github.com/terraform-google-modules/terraform-example-foundation/blob/master/docs/TROUBLESHOOTING.md#default-branch-setting ."
+        ERRORS+=$'git default branch must be configured.\n'
+    fi
+}
+
+# Validate the Configuration of the Gcloud CLI
```
bharathkkb, line 128:
```suggestion
# Validate the configuration of the Gcloud CLI
```

scripts/validate-requirements.sh
```diff
@@ -0,0 +1,274 @@
...
+
+    APPLICATION_DEFAULT_CREDENTIAL_OUTPUT="$(gcloud auth application-default print-access-token 2>&1 >/dev/null)"
+    if [ "$(echo "$APPLICATION_DEFAULT_CREDENTIAL_OUTPUT" | grep -c 'Could not automatically determine credentials')" -eq 1 ]; then
+        echo "You must configure an Application Default Credential."
+        echo "Visit https://cloud.google.com/sdk/gcloud/reference/auth/application-default/login and follow the instructions to authorize gcloud to access the Cloud Platform with Google user credentials."
+        ERRORS+=$'gcloud application default credential not configured.\n'
+    fi
+}
+
+# Function to validate the Configuration of the Gcloud CLI
```
bharathkkb, line 146:
```suggestion
# Function to validate the roles attached to credentialled account
```

scripts/validate-requirements.sh
```diff
@@ -0,0 +1,274 @@
...
+function validate_credential_roles(){
+    check_org_level_roles "$END_USER_CREDENTIAL" "$ORGANIZATION_ID"
+    check_billing_account_roles "$END_USER_CREDENTIAL" "$BILLING_ACCOUNT"
+}
+
+# Verifies wheter a user is assigned to the expected Orgzanization Level roles
```
bharathkkb, line 152:
```suggestion
# Verifies whether a user has the expected Organization level roles
```

scripts/validate-requirements.sh
```diff
@@ -0,0 +1,274 @@
...
+function check_org_level_roles(){
+
+    ORG_LEVEL_ROLES_OUTPUT=$(
+        gcloud organizations get-iam-policy "$2" \
+        --filter="bindings.members:$1" \
+        --flatten="bindings[].members" \
+        --format="table(bindings.role)" 2>/dev/null)
+
+    lines=$(echo "$ORG_LEVEL_ROLES_OUTPUT" | grep -o -e roles/resourcemanager.folderCreator -e roles/resourcemanager.organizationAdmin -e roles/orgpolicy.policyAdmin | wc -l)
```
bharathkkb, line 161: Can we move expected roles to a bash array so we can keep them separate from the function?
Also would using jq be simpler here?

scripts/validate-requirements.sh
```diff
@@ -0,0 +1,274 @@
...
+
+    if [ "$lines" -ne 3 ]; then
+        echo "The User must have the Organization Roles resourcemanager.folderCreator, resourcemanager.organizationAdmin and roles/orgpolicy.policyAdmin"
+        ERRORS+=$'There are missing organization level roles on the Credential.\n'
+    fi
+}
+
+# Verifies wheter a user is assigned to the expected Billing Level roles
```
bharathkkb, line 169:
```suggestion
# Verifies whether a user has the expected Billing Level roles
```

scripts/validate-requirements.sh
```diff
@@ -0,0 +1,274 @@
...
+        echo "The User must have the Billing Account Role billing.admin"
+        ERRORS+=$'There are missing billing account level roles on the Credential.\n'
+    fi
+}
+
+# Checks if initial config was done for 0-bootstrap step
+function validate_bootstrap_step(){
+    FILE=0-bootstrap/terraform.tfvars
+    if [ ! -f "$FILE" ]; then
+        echo "$FILE does not exists."
```
bharathkkb, line 190: Shouldn't this as the user to also rename the .example.tfvars to .tfvars?

scripts/validate-requirements.sh
```diff
@@ -0,0 +1,274 @@
...
+    else
+        if [ "$(grep -c REPLACE_ME $FILE)" != 0 ]; then
+            echo "$FILE must have required values fullfiled."
```
bharathkkb, line 193:
```suggestion
            echo "$FILE has have required values that must be replaced."
```

scripts/validate-requirements.sh
bharathkkb: Here and thoughout.

bharathkkb (commented):
Thanks for the PR @mauro-cit

scripts/validate-requirements.sh
mauro-cit: I've tried to follow the same idea we have for some CFT scripts, for example the one that realizes the lint tests (make docker_test_lint) so the user knows exactly what is missing on a single run. 
I've just added an exit for JQ for example and now when gcloud is missing, it doesn't run the validation for gcloud local config neither roles check (would fail anyway), so it makes the errors output way shorter, basically will only aggregate when is about Terraform, git and gcloud installation issues.

scripts/validate-requirements.sh
mauro-cit: @bharathkkb sorry but I couldn't get the idea, could you give a more descriptive comment about the issue?

scripts/validate-requirements.sh
mauro-cit: @bharathkkb I moved the roles to a specific constant variable, and now the checking if more dynamic. 
I couldn't visualize how JQ could help me better at this process. I think the new way I did is even simplier.

scripts/validate-requirements.sh
mauro-cit: We are looking at the version just like we have described them on the Readme. Then the compare _version function will consider every  "." separator. So, yes.

scripts/validate-requirements.sh
bharathkkb: sgtm

scripts/validate-requirements.sh
```diff
@@ -0,0 +1,331 @@
...
+    else
+        GIT_CURRENT_VERSION=$(git version | awk '{print $3}')
+        if [ "$(compare_version "$GIT_CURRENT_VERSION" "$GIT_VERSION")" -eq 2 ]; then
+            echo_wrong_version "git" "at least" "$GIT_VERSION" "https://git-scm.com/book/en/v2/Getting-Started-Installing-Git"
+            ERRORS+=$'git version is incompatible.\n'
+        fi
+    fi
+
+    if ! git config init.defaultBranch | grep "main" >/dev/null ; then
+        echo "git defaul branch must be configured."
```
bharathkkb, line 125:
```suggestion
        echo "git default branch must be configured."
```

scripts/validate-requirements.sh
```diff
@@ -0,0 +1,331 @@
...
+        echo "The User must have the Organization Roles resourcemanager.folderCreator, resourcemanager.organizationAdmin and roles/orgpolicy.policyAdmin"
+        ERRORS+=$'There are missing organization level roles on the Credential.\n'
+    fi
+}
+
+# Verifies whether a user has the expected Billing Level roles
+function check_billing_account_roles(){
+
+    BILLING_LEVEL_ROLES_OUTPUT=$(
+        gcloud beta billing accounts get-iam-policy "$2" \
```
bharathkkb, line 190: Looks like this may hang without warning if currently set project does not have `cloudbilling.googleapis.com` enabled. Can we require a project to be set and `cloudbilling.googleapis.com` enabled?

scripts/validate-requirements.sh
```diff
@@ -0,0 +1,331 @@
...
+        --filter="bindings.members:$1" \
+        --flatten="bindings[].members" \
+        --format="table(bindings.role)" 2>/dev/null)
```
bharathkkb, line 193: Isn't value better?
```suggestion
        --format="value(bindings.role)" 2>/dev/null)
```

scripts/validate-requirements.sh
bharathkkb: Hmm I tried out on my local machine and it failed with TF v 1.2.0

```
Validating Terraform installation...
An incompatible Terraform version was found.
Version required is exactly 0.13.7

```

```
{
  "terraform_version": "1.2.0",
  "platform": "darwin_amd64",
  "provider_selections": {},
  "terraform_outdated": true
}
```

mauro-cit: Yes, and it should. Currently the expected behavior of the validation is:
- Terraform must be exact version.
- Gcloud and Git must be at least a specific version. 

So if you don't have Terraform 0.13.7 (currently configured), it will fail anyway. 
Should we change this?

scripts/validate-requirements.sh
mauro-cit: So @bharathkkb I tested running this command with no default core/project set and it didn't raise any error/warning. 

Then I tested it with a default core/project but with the API cloudbilling.google.apis.com disabled and it didn't raise any error/warning as well. 

Is this the scenario you're refering to? If so, I couldn't reproduce the expected mentioned error.

scripts/validate-requirements.sh
bharathkkb: Let's pin to 1.0 now since https://github.com/terraform-google-modules/terraform-example-foundation/blob/015fe3d5716e9210cc45af9be73ae673e58df5fa/0-bootstrap/cb.tf#L20. It would be good if we can use that local as the source of truth.

scripts/validate-requirements.sh
bharathkkb: We can follow up if we get reports of this again. Thanks for investigating.

bharathkkb (approved):

bharathkkb:
@mauro-cit It will be good to follow up with updating docs to let users know about the script.

PR #784, opened 2022-08-26
Feat/ingress egress support for vpc sc

maxi-cit:
This adds support to ingress and egress rules within the perimeter in restricted vpc at 3-Network step, both on dual-svpc and hub-and-spoke


3-networks-dual-svpc/envs/development/main.tf
```diff
@@ -73,6 +73,8 @@ module "base_env" {
   default_region2                    = local.default_region2
   domain                             = var.domain
   parent_folder                      = var.parent_folder
+  ingress_policies                   = var.ingress_policies
+  egress_policies                    = var.egress_policies
```
bharathkkb, line 77: Let's directly set them here as opposed to using a var.

daniel-cit: Hi @bharathkkb  do you mean?
```
  ingress_policies                   = []
  egress_policies                    = []
```

maxi-cit: Hi Barath, variables were added there to support some assertions on integrations tests. Should I still delete them?

bharathkkb: Sounds good, lets keep for now. Looks like also need a rebase.

bharathkkb (approved):

PR #788, opened 2022-08-31
0-bootstrap fixes

daniel-cit:
0-bootstrap fixes:
- Typos
- error messages in validation scripts
- links


scripts/validate-requirements.sh
```diff
@@ -235,9 +235,9 @@ function echo_missing_installation () {
 # $3 = version to be displayed
 # $4 = web site to find the installation details of the missing binary
 function echo_wrong_version () {
```
mauro-cit, line 237: As we added a new parameter **$5**, it would be good to comment it's usage here as well.

daniel-cit: Good catch :)  
Fixed

felipecrescencio-cit (approved):
lgtm

1-org/README.md
```diff
@@ -115,7 +115,6 @@ Run `terraform output cloudbuild_project_id` in the `0-bootstrap` folder to see
    adjust your copy paths accordingly.
    ```
    cd gcp-policies
-   git checkout -b main
```
bharathkkb, line 118: Don't we need this if the repo is empty?

bharathkkb (approved):

1-org/README.md
daniel-cit: if the [default branch](https://github.com/terraform-google-modules/terraform-example-foundation/blob/master/docs/TROUBLESHOOTING.md#default-branch-setting) is set to main we don't need it.

I checked and it still works to run `git checkout -b main` if the default branch is already set to `main`.

 I will revert this change so that it will work if the default branch is main or master

bharathkkb (approved):

PR #791, opened 2022-09-01
Feature/new org policies

mauro-cit:
Adding the following new Organization Policies:

- Require OS Login: this one was already present, but now it's enabled by default
- Restrict Protocol Forwarding Based on type of IP Address
- Disable VPC External IPv6 usage
- Disable Service Account Key Upload
- Domain restricted contacts
- Sets the internal DNS setting for new projects to Zonal DNS Only


1-org/envs/shared/variables.tf
```diff
@@ -317,3 +317,9 @@ variable "backend_bucket" {
   description = "Backend bucket to load remote state information from previous steps."
   type        = string
 }
+
+variable "essential_contacts_domains_to_allow" {
+  description = "The list of domains that email addresses added to Essential Contacts can have."
+  type        = list(string)
+  default     = ["@clsecteam.com"]
```
daniel-cit, line 324: this variable should have no default

1-org/envs/shared/variables.tf
```diff
@@ -321,5 +321,4 @@ variable "backend_bucket" {
 variable "essential_contacts_domains_to_allow" {
   description = "The list of domains that email addresses added to Essential Contacts can have."
   type        = list(string)
-  default     = ["@clsecteam.com"]
```
daniel-cit: it is necessary to update the inputs in the readme file: `make docker_generate_docs`

1-org/envs/shared/README.md
```diff
@@ -18,7 +18,8 @@
 | dns\_hub\_project\_budget\_amount | The amount to use as the budget for the DNS hub project. | `number` | `1000` | no |
 | domains\_to\_allow | The list of domains to allow users from in IAM. Used by Domain Restricted Sharing Organization Policy. Must include the domain of the organization you are deploying the foundation. To add other domains you must also grant access to these domains to the terraform service account used in the deploy. | `list(string)` | n/a | yes |
 | enable\_hub\_and\_spoke | Enable Hub-and-Spoke architecture. | `bool` | `false` | no |
-| enable\_os\_login\_policy | Enable OS Login Organization Policy. | `bool` | `false` | no |
+| enable\_os\_login\_policy | Enable OS Login Organization Policy. | `bool` | `true` | no |
```
bharathkkb, line 21: Let's not expose this as a var anymore unless there is user demand.

1-org/envs/shared/org_policy.tf
```diff
@@ -103,6 +103,40 @@ module "org_shared_require_os_login" {
...
+
+module "internal_dns_on_new_project_to_zonal_dns_only" {
+  source          = "terraform-google-modules/org-policy/google"
+  version         = "~> 5.1"
+  organization_id = local.organization_id
+  folder_id       = local.folder_id
+  policy_for      = local.policy_for
+  policy_type     = "boolean"
+  enforce         = "true"
+  constraint      = "constraints/compute.setNewProjectDefaultToZonalDNSOnly"
```
bharathkkb, line 137: nit: can we switch these to a module for_each to be more succinct?

1-org/envs/shared/terraform.example.tfvars
```diff
@@ -17,6 +17,10 @@
 // Must include the domain of the organization you are deploying the foundation.
 domains_to_allow = ["example.com"]
 
+// The list of domains that email addresses added to Essential Contacts can have.
+// The domains entered here must starts with "@"
```
bharathkkb, line 21: This seems like something we can do within the config rather than asking users to prefix?

test/setup/outputs.tf
```diff
@@ -76,6 +76,10 @@ output "domains_to_allow" {
   value = tolist([var.domain_to_allow])
 }
 
+output "essential_contacts_domains_to_allow" {
+  value = tolist(["@${var.domain_to_allow}"])
```
bharathkkb, line 80: Don't we have a separate var.essential_contacts_domains_to_allow?

bharathkkb (commented):
Thanks for the PR @mauro-cit

test/setup/outputs.tf
mauro-cit: This is just a matter of making the build tests from the pipeline simpler to be configured. We do kind of the samething for the google groups by providing the same variable "group_email" for all of them.

1-org/envs/shared/README.md
mauro-cit: This was something that Andy asked on the meeting where he presented us these new org policies. 

The requested change for this partitular "os_login_policy" was for us to swap this variable default value to be "true" (enabled) and then show the user that he can change it to false in case he fell into one of the limitations situations that this constraint might bring up to a particular scenario.

Could you please double check wheter we should keep or not this var for this case?

1-org/envs/shared/org_policy.tf
```diff
@@ -18,45 +18,68 @@ locals {
   organization_id = local.parent_folder != "" ? null : local.org_id
   folder_id       = local.parent_folder != "" ? local.parent_folder : null
   policy_for      = local.parent_folder != "" ? "folder" : "organization"
+  essential_contacts_domains_to_allow = concat(
+    [for domain in var.essential_contacts_domains_to_allow : "${domain}" if can(regex("^@.*$", domain)) == true],
+    [for domain in var.essential_contacts_domains_to_allow : "@${domain}" if can(regex("^@.*$", domain)) == false]
+  )
+  boolean_type_organization_policies = {
+    org_disable_nested_virtualization = {
```
bharathkkb, line 26: Do we use the key anywhere? I was thinking this could be a set instead.

test/setup/outputs.tf
bharathkkb: Ah, I thought this was in the 1-org output, thanks for clarifying.

PR #799, opened 2022-09-08
create billing dataset in multi-regional by default

vponnam:
Creates billing dataset in the multi-regional mode as it's a [requirement](https://cloud.google.com/billing/docs/how-to/export-data-bigquery#:~:text=When%20exporting%20detailed%20usage%20cost%20data%2C%20you,to%20the%20appropriate%20tables%20within%20that%20dataset.) to be able to export billing data.

Users realize the multi-regional dataset limitation post running into an error, and this addressed the need for recreating the dataset.


bharathkkb (approved):
Thanks for the fix @vponnam

PR #818, opened 2022-09-16
Grant role browser to the terraform service account for running gcloud beta terraform vet

daniel-cit:
Grant the role Browser to the terraform service accounts create in step 0-bootstrap so that  the service account can read the ancestry of folders when running gcloud beta terraform vet.


0-bootstrap/sa.tf
```diff
@@ -35,14 +35,20 @@ locals {
       "roles/resourcemanager.organizationViewer",
       "roles/accesscontextmanager.policyAdmin",
       "roles/essentialcontacts.admin",
+      "roles/browser",
+    ],
+    "env" = [
+      "roles/browser",
```
bharathkkb, line 41: Let's move this to a common_roles local and then concat local here. Then for the common_roles we can add the justification of why we added this and the user can remove if needed.

bharathkkb (approved):
Minor nit

0-bootstrap/sa.tf
daniel-cit: `common_roles` local created with an explanation for the need of `browser` role

bharathkkb (approved):

PR #826, opened 2022-09-29
Retry on transient errors

daniel-cit:
Retry once on possible transient errors:

- Rate limit exceeded
- Failed to delete non empty folder
- Permission denied on resource


bharathkkb (approved):
Thanks @daniel-cit!

PR #838, opened 2022-10-04
Review builds with Jenkins

felipecrescencio-cit:
Reviewed executing all steps using Jenkins instead of Cloud Build:
- Fixed instructions
- Fixed Jenkins Agent setup script to install wget to download terraform and install terraform-tools using apt because gcloud install is disabled in default image
- Bumped Jenkins Agent image to Debian 11 - new GCE default
- Add flag to disable App Infra Pipeline when using Jenkins deploy


0-bootstrap/README-Jenkins.md
```diff
@@ -202,7 +202,6 @@ You arrived to these instructions because you are using the `jenkins_bootstrap`
 
 1. Run terraform commands.
    - After the credentials are configured, we will create the `prj-b-seed` project (which contains the GCS state bucket and Terraform custom service account) and the `prj-b-cicd` project (which contains the Jenkins Agent, its custom service account and where we will add VPN configuration)
-   - **WARNING: Make sure you have commented-out the `cloudbuild_bootstrap` module and enabled the `jenkins_bootstrap` module in the `./main.tf` file**
```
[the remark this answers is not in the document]
felipecrescencio-cit: @daniel-cit I think it is already mentioned here:
https://github.com/felipecrescencio-cit/terraform-example-foundation/blob/review-jenkins/0-bootstrap/README-Jenkins.md?plain=1#L143-L158

4-projects/business_unit_2/shared/variables.tf
```diff
@@ -42,3 +42,9 @@ variable "remote_state_bucket" {
   description = "Backend bucket to load Terraform Remote State Data from previous steps."
   type        = string
 }
+
+variable "enable_cloudbuild_deploy" {
```
bharathkkb, line 46: Rather than exposing a var, can we use one of the [jenkins output ](https://github.com/terraform-google-modules/terraform-example-foundation/blob/253de5e03edf8105ba9262758438e1e6777d494a/0-bootstrap/outputs.tf#L119)from bootstrap to simplify this?

bharathkkb (commented):
Can we add an initial disclaimer in the jenkins readme that Jenkins support will be deprecated in a future release.

4-projects/business_unit_2/shared/variables.tf
felipecrescencio-cit: Done

1-org/.gitignore
```diff
@@ -28,6 +28,9 @@ Session.vim
 *.tfstate
 *.tfstate.*
 
+# tf lock file
+**/.terraform.lock.hcl
```
bharathkkb, line 32: We should allow checking in lockfiles into each stage consumer git repo. Could we create a global .gitignore at repo root so we do check it into this repo?

bharathkkb (approved):

1-org/.gitignore
felipecrescencio-cit: Done

PR #853, opened 2022-10-13
remove unecessary variables in network-dual-svpc/shared

guissalustiano:
On run `grep` to show variables used in folder I realize that not always in declaration is used
image

This PR remove unnecessary variable
(You can check that has 9 variables on grep, and 9 in `variable.tf` now)


daniel-cit:
@guissalustiano thanks for the pull request.

the cause of the failed build is:

```
plan.go:34: 
Error Trace: /workspace/test/integration/shared/plan.go:34
             /workspace/test/integration/shared/terraform.go:384
             /workspace/test/integration/shared/terraform.go:396
             /workspace/test/integration/shared/terraform.go:408
             /workspace/test/integration/shared/terraform.go:432
             /workspace/test/integration/shared/stages.go:31
             /workspace/test/integration/shared/terraform.go:432
Error: Received unexpected error:
       FatalError{Underlying: error while running command: exit status 1; 
       Error: Value for undeclared variable
       
       A variable named "access_context_manager_policy_id" was assigned on the
       command line, but the root module does not declare a variable of that name.
       To use this value, add a "variable" block to the configuration.
       
       Error: Value for undeclared variable
       
       A variable named "terraform_service_account" was assigned on the command
       line, but the root module does not declare a variable of that name. To use
       this value, add a "variable" block to the configuration.
       }
Test:       	TestShared
2022/10/14 01:52:32 RUN_STAGE env var set to apply
2022/10/14 01:52:32 Skipping stage teardown
--- FAIL: TestShared (12.29s)
```

the cause is probably that with this change you will need to have two sets of variables in the test, one for the [Shared VPC](https://cloud.google.com/architecture/security-foundations/networking#vpcsharedvpc-id7-1-shared-vpc-) network mode and one for the [Hub-and-spoke](https://cloud.google.com/architecture/security-foundations/networking#hub-and-spoke) network mode:

https://github.com/terraform-google-modules/terraform-example-foundation/blob/4ade8c17dcf20b8553a49d9984ff1ca922ffa3cc/test/integration/shared/shared_test.go#L55-L59

like is done for the `tfDir` option

https://github.com/terraform-google-modules/terraform-example-foundation/blob/4ade8c17dcf20b8553a49d9984ff1ca922ffa3cc/test/integration/shared/shared_test.go#L66-L70

bharathkkb (approved):
Thanks for the PR @guissalustiano

PR #859, opened 2022-10-22
fix the relative path used in the chdir param of terraforrm state command

daniel-cit:
Fix the relative path used in the `-chdir` param of `terraforrm state` command to assume that the user is executing commands at the same level of the `terraform-example-foundation` directory before cloning the Google Cloud Source Repositories.


bharathkkb (approved):
Thanks for the fix

PR #863, opened 2022-10-26
Add support for new organization policies

daniel-cit:
This pull request add support for the following organization policies:

- Enforce Public Access Prevention

  >Secure your Cloud Storage data from public exposure by enforcing public access prevention.
  >This governance policy prevents existing and future resources from being accessed via the public internet by disabling and blocking ACLs and IAM permissions that grant access to allUsers and allAuthenticatedUsers.
  >Enforce this policy on the entire organization (recommended), specific projects, or specific folders to ensure no data is publicly exposed.
  >This policy overrides existing public permissions.
  >Public access will be revoked for existing buckets and objects after this policy is enabled.

- Restrict Authorized Networks on Cloud SQL instances

  >This boolean constraint restricts adding Authorized Networks for unproxied database access to Cloud SQL instances where this constraint is set to True.
  >This constraint is not retroactive, Cloud SQL instances with existing Authorized Networks will still work even after this constraint is enforced.
  >By default, Authorized Networks can be added to Cloud SQL instances.


bharathkkb (commented):
Thanks @daniel-cit Can we update the tests too https://github.com/terraform-google-modules/terraform-example-foundation/blob/a0604b3aff804e2c33c6d936c72d356c263bfcb1/test/integration/org/org_test.go#L137

daniel-cit:
> Thanks @daniel-cit Can we update the tests too
> 
> https://github.com/terraform-google-modules/terraform-example-foundation/blob/a0604b3aff804e2c33c6d936c72d356c263bfcb1/test/integration/org/org_test.go#L137

good catch!

New org policies added to the test too

bharathkkb (approved):

PR #891, opened 2022-11-28
feat!: use random_project_id_length

apeabody:


bharathkkb (approved):
Thanks @apeabody! Could we also add `random_project_id_length` in other places to prevent collisions?

https://github.com/search?q=repo%3Aterraform-google-modules%2Fterraform-example-foundation%20random_project_id&type=code

apeabody:
> 

Sure!

daniel-cit:
Many projects are at the limit or near the limit of the max `project_id` length (30 chars) like:

```text
prj-bu1-n-sample-peering-aaaa
prj-bu1-n-sample-floating-aaaa
prj-bu1-n-sample-restrict-aaaa
prj-bu2-c-infra-pipeline-aaaa
prj-c-restricted-net-hub-aaaa
```

with four additional chars for the random suffix, they would brake the limit .

Not all the projects can have this increase in the random size to prevent collisions

apeabody:
> Many projects are at the limit or near the limit of the max `project_id` length (30 chars) like:
> 
> ```
> prj-bu1-n-sample-peering-aaaa
> prj-bu1-n-sample-floating-aaaa
> prj-bu1-n-sample-restrict-aaaa
> prj-bu2-c-infra-pipeline-aaaa
> prj-c-restricted-net-hub-aaaa
> ```
> 
> with four additional chars for the random suffix, they would brake the limit .
> 
> Not all the projects can have this increase in the random size to prevent collisions

Thanks for the feedback @daniel-cit.  These should all be currently using `random_project_id = true` which already adds 4 characters for randomness, so adding the `random_project_id_length = 4` should maintain the same length.  The primary difference is that `random_project_id_length` implements a different and larger collusion domain (https://github.com/terraform-google-modules/terraform-google-project-factory/pull/735), which should reduce duplicate `project_id` during CI.  That said, let me know if I overlooked anything.  Cheers!

daniel-cit:
> > Many projects are at the limit or near the limit of the max `project_id` length (30 chars) like:
> > ```
> > prj-bu1-n-sample-peering-aaaa
> > prj-bu1-n-sample-floating-aaaa
> > prj-bu1-n-sample-restrict-aaaa
> > prj-bu2-c-infra-pipeline-aaaa
> > prj-c-restricted-net-hub-aaaa
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > with four additional chars for the random suffix, they would brake the limit .
> > Not all the projects can have this increase in the random size to prevent collisions
> 
> Thanks for the feedback @daniel-cit. These should all be currently using `random_project_id = true` which already adds 4 characters for randomness, so adding the `random_project_id_length = 4` should maintain the same length. The primary difference is that `random_project_id_length` implements a different and larger collusion domain ([terraform-google-modules/terraform-google-project-factory#735](https://github.com/terraform-google-modules/terraform-google-project-factory/pull/735)), which should reduce duplicate `project_id` during CI. That said, let me know if I overlooked anything. Cheers!

I see. Thanks for the clarification.
This is a great improvement ❤️ 

A great  follow up would be adding the `random_project_id_length`  to the two projects from 0-bootstrap that are created by the bootstrap modules

- https://github.com/terraform-google-modules/terraform-google-bootstrap/blob/ce421d96b85bb4c6b5475e582f2fb102abd50191/main.tf#L59
- https://github.com/terraform-google-modules/terraform-google-bootstrap/blob/ce421d96b85bb4c6b5475e582f2fb102abd50191/modules/tf_cloudbuild_source/main.tf#L33

apeabody:
> terraform-google-modules/terraform-google-project-factory#735



> > > Many projects are at the limit or near the limit of the max `project_id` length (30 chars) like:
> > > ```
> > > prj-bu1-n-sample-peering-aaaa
> > > prj-bu1-n-sample-floating-aaaa
> > > prj-bu1-n-sample-restrict-aaaa
> > > prj-bu2-c-infra-pipeline-aaaa
> > > prj-c-restricted-net-hub-aaaa
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > with four additional chars for the random suffix, they would brake the limit .
> > > Not all the projects can have this increase in the random size to prevent collisions
> > 
> > 
> > Thanks for the feedback @daniel-cit. These should all be currently using `random_project_id = true` which already adds 4 characters for randomness, so adding the `random_project_id_length = 4` should maintain the same length. The primary difference is that `random_project_id_length` implements a different and larger collusion domain ([terraform-google-modules/terraform-google-project-factory#735](https://github.com/terraform-google-modules/terraform-google-project-factory/pull/735)), which should reduce duplicate `project_id` during CI. That said, let me know if I overlooked anything. Cheers!
> 
> I see. Thanks for the clarification. This is a great improvement ❤️
> 
> A great follow up would be adding the `random_project_id_length` to the two projects from 0-bootstrap that are created by the bootstrap modules
> 
> * https://github.com/terraform-google-modules/terraform-google-bootstrap/blob/ce421d96b85bb4c6b5475e582f2fb102abd50191/main.tf#L59
> * https://github.com/terraform-google-modules/terraform-google-bootstrap/blob/ce421d96b85bb4c6b5475e582f2fb102abd50191/modules/tf_cloudbuild_source/main.tf#L33

We can think about it, but would be a breaking change and need a migration guide for existing project_id: https://github.com/terraform-google-modules/terraform-google-bootstrap/pull/215

PR #893, opened 2022-12-01
enable missing DNS logging

daniel-cit:
Enable missing DNS logging:
- Jenkins network
- Cloud Build peering network
- 4-projects peering example

### DIFF
--- a/0-bootstrap/modules/cb-private-pool/network.tf
+++ b/0-bootstrap/modules/cb-private-pool/network.tf
@@ -29,7 +29,7 @@ module "peered_network" {
 
   subnets = [
     {
-      subnet_name           = "sb-b-peer-${var.private_worker_pool.region}"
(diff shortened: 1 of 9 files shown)


0-bootstrap/modules/cb-private-pool/network.tf
```diff
@@ -40,6 +40,18 @@ module "peered_network" {
 
 }
 
+resource "google_dns_policy" "default_policy" {
+  count = var.private_worker_pool.create_peered_network ? 1 : 0
+
+  project                   = var.project_id
+  name                      = "dp-b-peer-default-policy"
```
bharathkkb, line 47: Would it be better to signify this is for private pools?
```suggestion
  name                      = "dp-b-cbpools-default-policy"
```

daniel-cit: fixed

bharathkkb (approved):

PR #895, opened 2022-12-01
Fix source repos keys in sa_roles map

felipecrescencio-cit:
Fix source repos to allow user to create as many applications workspaces as user needs and just provide the roles (sa_roles) to user applications.
Previous to this fix user will need to provide each repo key for all applications in sa_roles map.

### DIFF
--- a/4-projects/modules/single_project/main.tf
+++ b/4-projects/modules/single_project/main.tf
@@ -17,7 +17,10 @@
 locals {
   env_code        = element(split("", var.environment), 0)
   shared_vpc_mode = var.enable_hub_and_spoke ? "-spoke" : ""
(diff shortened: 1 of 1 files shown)


bharathkkb (commented):
Thanks for the PR @felipecrescencio-cit 
I did not fully understand what you meant by `Previous to this fix user will need to provide each repo key for all applications in sa_roles map.`.  Could you clarify the usecase?

bharathkkb (approved):
Synced offline. This is due to the fact that `app_infra_pipeline_service_accounts` keys could have more than one repo which may not be present in the sa_roles map leading to a error when looking up `var.sa_roles[{local.source_repos.value}]`.

PR #896, opened 2022-12-01
remove Editor from Seed and CICD projects

daniel-cit:
remove Editor role  from Seed and CICD projects in 0-bootrstrap

### DIFF
--- a/0-bootstrap/modules/parent-iam-remove-role/variables.tf
+++ b/0-bootstrap/modules/parent-iam-remove-role/variables.tf
@@ -1,0 +1,35 @@
+/**
+ * Copyright 2022 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+variable "parent_type" {
+  description = "Type of the parent resource. valid values are `organization`, `folder`, and `project`."
+  type        = string
+
+  validation {
+    condition     = var.parent_type == "organization" || var.parent_type == "folder" || var.parent_type == "project"
+    error_message = "For parent_type only `organization`, `folder`, and `project` are valid."
+  }
+}
+
+variable "parent_id" {
+  description = "ID of the parent resource."
+  type        = string
+}
+
+variable "roles" {
+  description = "Roles to remove all members in the parent resource."
+  type        = list(string)
+}
--- a/0-bootstrap/modules/parent-iam-remove-role/variables.tf
+++ b/0-bootstrap/modules/parent-iam-remove-role/variables.tf
@@ -19,7 +19,7 @@ variable "parent_type" {
   type        = string
 
   validation {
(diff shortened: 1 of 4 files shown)


0-bootstrap/modules/parent-iam-remove-role/variables.tf
```diff
@@ -0,0 +1,35 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+variable "parent_type" {
+  description = "Type of the parent resource. valid values are `organization`, `folder`, and `project`."
+  type        = string
+
+  validation {
+    condition     = var.parent_type == "organization" || var.parent_type == "folder" || var.parent_type == "project"
```
bharathkkb, line 22: nit: might be simpler
```suggestion
    condition     = contains(["organization", "folder", "project"], var.parent_type)
```

0-bootstrap/modules/parent-iam-remove-role/versions.tf
```diff
@@ -0,0 +1,32 @@
...
+      source  = "hashicorp/google"
+      version = ">= 3.77"
+    }
+    google-beta = {
+      source  = "hashicorp/google-beta"
+      version = ">= 3.77"
+    }
+    random = {
+      source = "hashicorp/random"
+    }
```
bharathkkb, line 30: Doesn't look like we use both of these?

0-bootstrap/sa.tf
```diff
@@ -175,6 +175,31 @@ module "cicd_project_iam_member" {
   roles       = each.value
 }
 
+module "seed_project_remove_editor" {
+  source = "./modules/parent-iam-remove-role"
+
+  parent_type = "project"
+  parent_id   = module.seed_bootstrap.seed_project_id
```
bharathkkb, line 182: nit: Does for_each throw an error due to computed project_id?

0-bootstrap/sa.tf
```diff
@@ -175,6 +175,31 @@ module "cicd_project_iam_member" {
   roles       = each.value
 }
 
+module "seed_project_remove_editor" {
```
bharathkkb, line 178: It would be good to add a comment as a rationale.

bharathkkb (commented):
Thanks @daniel-cit

0-bootstrap/sa.tf
daniel-cit: we can use a `for_each` if we use the map option.
Good Catch.

0-bootstrap/sa.tf
daniel-cit: Comment added

0-bootstrap/modules/parent-iam-remove-role/variables.tf
daniel-cit: Fixed. 
Thanks for the suggestion.

0-bootstrap/sa.tf
```diff
@@ -176,6 +181,25 @@ module "cicd_project_iam_member" {
   roles       = each.value
 }
 
+// In the bootstrap projects is created, the Compute Engine
```
bharathkkb, line 184:
```suggestion
// When the bootstrap projects are created, the Compute Engine
```

0-bootstrap/sa.tf
```diff
@@ -176,6 +181,25 @@ module "cicd_project_iam_member" {
...
+// default service account is disabled but it still has the Editor
+// role associated with the service account. When the projects are created
+// the Compute Engine default service account is the only Editor.
```
bharathkkb, line 187:
```suggestion
// role associated with the service account. This default SA is the
// only member with the editor role.
```

bharathkkb (approved):

PR #907, opened 2022-12-07
fix: remove locals related to hub and spoke from dual shared vpc code

### LINKED ISSUE #871: Remove unused code on 3-networks-dual-svpc
Remove these [code lines](https://github.com/terraform-google-modules/terraform-example-foundation/blob/40e391c0edb3650f3a5a940062194e4b12add1b7/3-networks-dual-svpc/modules/base_env/main.tf#L33-L42) from step 3-networks-dual-svpc base_env main.
These locals are not used in this step.

daniel-cit:
fixes #871

### DIFF
--- a/3-networks-dual-svpc/modules/base_env/main.tf
+++ b/3-networks-dual-svpc/modules/base_env/main.tf
@@ -32,16 +32,6 @@ locals {
 
 
   bgp_asn_number = var.enable_partner_interconnect ? "16550" : "64514"
-  /*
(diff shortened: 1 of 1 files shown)


bharathkkb (approved):
Thanks @daniel-cit Can we also try bumping the devtools to 1.9 which has tflint (https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/pull/1272) and will catch these issues?

PR #921, opened 2023-01-17
fix: add cloud build bucket location

hnorm:
adds var.default_region as the location of the cloud build bucket


bharathkkb (approved):
Thanks for the fix!

PR #983, opened 2023-06-29
Added a script to remove example-foundation resources automatically.

fakeskimo:
I have found that it's difficult to remove all the resources after deploying example-foundation during practice and production. I also got same feedback from many of my colleagues and partners I work with. Therefore I made this script to remove all the resources deployed by terraform-example-foundation. I think it would be helpful to include this script for who want to learn terraform-example-foundation. Please review this script and let me know what is your thought. Thanks! - Junho

### DIFF
--- a/scripts/remove-example-foundation.sh
+++ b/scripts/remove-example-foundation.sh
@@ -1,0 +1,385 @@
+#!/usr/bin/env bash
+
+# Copyright 2022 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+################################################################################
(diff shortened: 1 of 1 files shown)


fmichaelobrien:
Reviewing...
This undeploy script is definitely required.  While we have terraform destroy - for tf managed resources - all gcloud additions need a corresponding removal that does not necessarily delete the bootstrap project - to keep quota within the 30d limit during repetitive development.

scripts/remove-example-foundation.sh
```diff
@@ -0,0 +1,385 @@
...
+FLDR_LIST=$(gcloud resource-manager folders list \
+    --organization "$ORG_NUM" --filter="displayName:fldr*" \
+    --format="value(name)")
+
+# Project list from the organization
+PJT_LIST=$(gcloud projects list --filter="projectId:prj-*" \
+    --format="value(projectId)")
+
+################################################################################
+
```
c4th4r5y5, line 82: I've found this script very useful for tearing down the example foundation :rocket:
 
I would like to contribute this block which checks the currently authenticated user for the Owner role on the organization. This might be too broad of a IAM permission but we could suggest removing it after running this script (or make the check more specific on the granualr IAM permissions): 

```shell
echo ""
echo "# 0. IAM check"

if echo "$(gcloud alpha organizations get-iam-policy $ORG_NUM )" | grep -q 'roles/owner'; then
    echo "You are properly configured with IAM."
else
    echo -e "Looks like you do not have the Owner role on organization($ORG_NUM)."
    exit 1
fi
```

PR #986, opened 2023-07-13
chore: add apply/destroy actions for all stages for the foundation deploy helper

daniel-cit:
This PR adds apply/destroy actions for all stages for the foundation deploy helper


amandakarina:
The message in https://github.com/terraform-google-modules/terraform-example-foundation/pull/986/files#diff-85ada0ac366ede812a636143928cd586cd2802c729d364267cec07e0e2b96aa3R69 can be re-write to inform the user that they can see the logs, but they don't need to wait to press enter.

bharathkkb (approved):

PR #991, opened 2023-08-17
feat!: Network Refactoring

Samir-Cit:
Hello folks.

This PR is to refactor the network resources into the new folder layout.

PR #992, opened 2023-08-29
feat: update tf-wrapper.sh script to deal with generic folder hierarchy

daniel-cit:
This PR updates the `tf-wrapper.sh` script to deal with generic source code folder hierarchy

depends on #991


build/tf-wrapper.sh
```diff
@@ -23,78 +23,189 @@ project_id=$4
...
+#   else
+#       if [[ "$tf_leaf" =~ $leaf_regex_action  && "$tf_root" == "shared" && "$branch" == "production" ]]; then
+#         echo "true"
+#       else
+#         echo "false"
+#       fi
+#   fi
+# }
+
+##### End of default source organization #####
```
amandakarina, line 129:
```suggestion
##### End of alternative source organization #####
```

docs/change_resource_hierarchy.md
```diff
@@ -1,168 +1,39 @@
...
 
-The current deployment scenario of Terraform Foundation Example blueprint considers a flat resource hierarchy where all folders are at the same level and have one folder for each environment. Here is a detailed explanation of each folder:
+The current deployment scenario of Terraform Foundation Example blueprint considers a flat resource hierarchy where all folders are at the same level and have one folder for each environment and three special folders. Here is a detailed explanation of each folder:
 
 | Folder | Description |
 | --- | --- |
 | bootstrap | Contains the seed and CI/CD projects that are used to deploy foundation components. |
-| common | Contains projects with common resources used by the organization like logging, SCC and hybrid connectivity. |
+| common | Contains projects with common resources used by the organization like logging and Security Command Center. |
+| network |  Contains projects with common networks resources used by the organization like DNS Hub, hybrid connectivity, and Shared VPCs. |
```
amandakarina, line 11:
```suggestion
| network | Contains projects with common networks resources used by the organization like DNS Hub, hybrid connectivity, and Shared VPCs. |
```

daniel-cit: done

build/tf-wrapper.sh
daniel-cit: done

build/tf-wrapper.sh
```diff
@@ -23,78 +23,189 @@ project_id=$4
...
 
-## terraform init/plan/validate for all valid environments matching regex.
+#============================================================================#
+# terraform init/plan/validate for all valid environments matching condition.
+#============================================================================#
 tf_plan_validate_all() {
-  local env
-  local component
+  local leaf
   find "$base_dir" -mindepth 1 -maxdepth 1 -type d \
```
jasonbisson, line 198: Should the mindepth & maxdepth be set to variables too?

build/tf-wrapper.sh
```diff
@@ -152,53 +265,53 @@ tf_validate() {
...
 
+#=================================================================#
 # Runs single action for each instance of env in folder hierarchy.
+#=================================================================#
 single_action_runner() {
-  local env
-  local component
+  local leaf
+  # filter folders that does not contain Terraform configurations
   find "$base_dir" -mindepth 1 -maxdepth 1 -type d \
```
jasonbisson, line 282: Same comment about the variable for mindepth and maxdepth variables.

jasonbisson (commented):
Just a couple of comments about the mindepth and maxdepth variables. Other than that it LGTM

build/tf-wrapper.sh
daniel-cit: Hi @jasonbisson , thanks for the review

The sole purpose of this code section and the other code section is to scan the base directory of the repository to filter directories that do not contain terraform configuration that should be considered for evaluation.

The result of this filter is the initial set of directories we would use as our base scan

For this code section, "mindepth" should always be one because we don’t want to miss any directory and "maxdepth" should always be one too so that the filtered list will always have the same depth.  if we had for example  "maxdepth" 2 we would have in the result list

- envs
- envs/shared

as starting point for the next step

This is an initial filter to trim directories that we know should not contain configurations for the environments.

build/tf-wrapper.sh
daniel-cit: same as the other code section

jasonbisson (approved):
LGTM

bharathkkb (approved):

PR #1015, opened 2023-11-07
feat: CAI Monitoring Cloud Function

Samir-Cit:
Hi folks.

This PR is to add a new Cloud Function that will monitor CAI assets with IAM Policy content type.

The Cloud Function will search for IAM members who has any role from a list of roles pre-defined.
If one role of the list of pre-defined roles is grant to a member, it will generate a SCC Finding with:
- Time that the role was granted
- Member who has the roles (user, service account, group...)
- Roles that was granted


1-org/modules/cai-monitoring/README.md
```diff
@@ -0,0 +1,71 @@
+# Cloud Asset Inventory Notification
+Uses Google Cloud Asset Inventory to create a feed of IAM Policy change events, then process them to detect when they include a gmail address and revert that IAM policy change using a Cloud Function.  The terraform directory contains all the code required to deploy the solution (including the cloud function).
```
ellioteaton, line 2: I think there's some leftover from the previous sample.
"detect when they include a gmail address" should be "detect when the IAM policy change grants certain sensitive roles (Owner, Editor, Organization Administrator, Network Administrator, Organization Firewall Policy Administrator)"

1-org/envs/shared/projects.tf
```diff
@@ -158,7 +158,7 @@ module "scc_notifications" {
   org_id                   = local.org_id
   billing_account          = local.billing_account
   folder_id                = google_folder.common.id
-  activate_apis            = ["logging.googleapis.com", "pubsub.googleapis.com", "securitycenter.googleapis.com", "billingbudgets.googleapis.com"]
+  activate_apis            = ["logging.googleapis.com", "pubsub.googleapis.com", "securitycenter.googleapis.com", "billingbudgets.googleapis.com", "cloudkms.googleapis.com"]
```
ellioteaton, line 161: For my understanding, are we adding new KMS features?
Or are the kms references just to demonstrate the code sample that the customer could use to enable CMEK encryption on the feed/function?

1-org/modules/cai-monitoring/README.md
Samir-Cit: Oh, it is... Soon I'll push this fix!

1-org/envs/shared/projects.tf
Samir-Cit: Yes, I'm adding new KMS Features to use on:
- **Artifact Registry**
- Source code **Bucket**
- **Pub/Sub**

But the _cai-monitoring_ module itself accepts a null value for _encryption_key_.

1-org/envs/shared/cai_monitoring.tf
```diff
@@ -0,0 +1,37 @@
...
+
+module "kms" {
+  source  = "terraform-google-modules/kms/google"
+  version = "~> 2.1"
+
+  project_id      = module.scc_notifications.project_id
+  keyring         = "krg-cai-monitoring"
+  location        = local.default_region
+  keys            = ["key-cai-monitoring"]
+  prevent_destroy = "false"
```
daniel-cit, line 25: there should be a input similar to https://github.com/terraform-google-modules/terraform-example-foundation/blob/a92e31b3c6e48ca1c13570e96c88042ec65d868a/1-org/envs/shared/variables.tf#L85C11-L85C43
to set this flag

daniel-cit: this flag also need to be added to the helper script:
- when reading the tfvars file 
- when writing the 1-org tfvars file
- in the validation that checks if the flags to destroy resources has been set

and to the 1-org vars map used  in the integration test to create the test

Samir-Cit:
!gcbrun

1-org/envs/shared/projects.tf
Samir-Cit: Today we have some KMS feature also on step 4-projects.
This KMS is to protect the resources that supports, like the ones that I listed above.

I added KMS feature in _cai-monitoring_ so the user can choose to use or not KMS protection.
But this module accepts a **null value** on **encryption_key** variable, and **true** on **enable_cmek** variable, that way it will not use KMS on the resources listed above.

1-org/envs/shared/variables.tf
```diff
@@ -202,3 +202,9 @@ variable "create_unique_tag_key" {
   type        = bool
   default     = false
 }
+
+variable "cai_monitoring_kms_force_destroy" {
+  description = "(Optional) If set to true, delete all contents when destroying the resource; otherwise, destroying the resource will fail if contents are present."
```
gtsorbo, line 207: Let's provide a more specific description since this applies only to the KMS resources
```suggestion
  description = "If set to true, delete KMS keyring and keys when destroying the module; otherwise, destroying the module will fail if KMS keys are present."
```

gtsorbo (changes requested):
small changes, but otherwise great

gtsorbo (approved):

PR #1025, opened 2023-11-13
feat!: Add support for Log Analytics and Remove BigQuery log destination

daniel-cit:
This PR adds support for [Log Analytics](https://cloud.google.com/logging/docs/log-analytics#analytics) in the Log Bucket destination and removes the  BigQuery log destination from the centralized logging configured in step `1-org`.

Now that Log Analytics is available in GA, the same use cases for analytics that were previously accomplished using a log sink export to BigQuery can now be accomplished with a log sink export to a Logging bucket. 

This change helps address use cases for both writing custom SQL queries for deep analysis analysis  and ad-hoc investigation using the Logs Explorer console, with one single log sink and data source instead of two. 

For more information on the differences of these log sink types see [Moving to Log Analytics for BigQuery export users](https://cloud.google.com/blog/products/data-analytics/moving-to-log-analytics-for-bigquery-export-users)


1-org/modules/centralized-logging/README.md
eeaton, line 1: Do we still need this if removing the BQ sink?

1-org/modules/centralized-logging/README.md
```diff
@@ -73,7 +75,7 @@ module "logging_logbucket" {
...
 |------|-------------|------|---------|:--------:|
 | bigquery\_options | Destination BigQuery options:<br>- dataset\_name: The name of the bigquery dataset to be created and used for log entries.<br>- logging\_sink\_name: The name of the log sink to be created.<br>- logging\_sink\_filter: The filter to apply when exporting logs. Only log entries that match the filter are exported. Default is "" which exports all logs.<br>- expiration\_days: Table expiration time. If null logs will never be deleted.<br>- partitioned\_tables: Options that affect sinks exporting data to BigQuery. use\_partitioned\_tables - (Required) Whether to use BigQuery's partition tables.<br>- delete\_contents\_on\_destroy: If set to true, delete all contained objects in the logging destination. | <pre>object({<br>    dataset_name               = optional(string, null)<br>    logging_sink_name          = optional(string, null)<br>    logging_sink_filter        = optional(string, "")<br>    expiration_days            = optional(number, null)<br>    partitioned_tables         = optional(bool, true)<br>    delete_contents_on_destroy = optional(bool, false)<br>  })</pre> | `null` | no |
```
eeaton, line 77: bigquery_options can be removed from the table of input variables because we're removing the bq sink, right?

1-org/modules/centralized-logging/README.md
```diff
@@ -87,6 +89,7 @@ module "logging_logbucket" {
 |------|-------------|
 | bigquery\_destination\_name | The resource name for the destination BigQuery. |
```
eeaton, line 90: is bigquery_destination_name still needed after removing the bq sink?

1-org/modules/centralized-logging/README.md
```diff
@@ -73,7 +75,7 @@ module "logging_logbucket" {
...
...on: The location of the log bucket. Default: global.<br>- enable\_analytics: Whether or not Log Analytics is enabled. A Log bucket with Log Analytics enabled can be queried in the Log Analytics page using SQL queries. Cannot be disabled once enabled.<br>- linked\_dataset\_id: The ID of the linked BigQuery dataset. A valid link dataset ID must only have alphanumeric characters and underscores within it and have up to 100 characters.<br>- linked\_dataset\_description: A use-friendly description of the linked BigQuery dataset. The maximum length of the description is 8000 characters.<br>- retention\_days: The number of days data should be retained for the log bucket. Default 30. | <pre>object({<br>    name                       = optional(string, null)<br>    logging_sink_name          = optional(string, null)<br>    logging_sink_filter        = optional(string, "")<br>    location                   = optional(string, "global")<br>    enable_analytics           = optional(bool, true)<br>    linked_dataset_id          = optional(string, null)<br>    linked_dataset_description = optional(string, null)<br>    retention_days             = optional(number, 30)<br>  })</pre> | `null` | no |
```
eeaton, line 78: Double checking... does linked_dataset_id allow you to specify a name for the linked dataset different from the name of the log bucket? 

I thought the linked dataset ID always took the name of the bucket because the gcloud command line doesn't have an option to specify a new dataset_id.
https://cloud.google.com/logging/docs/buckets#link-bq-dataset

(If this is just an inconsistency between the API and gcloud, then it's not a blocking issue)

daniel-cit: In Terraform the link id will be the name of the BigQuery dataset, I am not familiar with the gcloud command usage for creating the link

checking  In the logging API we have:

[Log bucket create](https://cloud.google.com/logging/docs/reference/v2/rest/v2/projects.locations.buckets.links/create#query-parameters) - **linkId**


```
Required. The ID to use for the link. The linkId can have up to 100 characters. A valid linkId must only have alphanumeric characters and underscores within it.
```


[Link.BigQueryDataset](https://cloud.google.com/logging/docs/reference/v2/rest/v2/billingAccounts.locations.buckets.links#Link.BigQueryDataset) - **datasetId**

```
Output only. The full resource name of the BigQuery dataset. The DATASET_ID will match the ID of the link, so the link must match the naming restrictions of BigQuery datasets (alphanumeric characters and underscores only).

The dataset will have a resource path of "bigquery.googleapis.com/projects/[PROJECT_ID]/datasets/[DATASET_ID]"
```

it is the same as in the [API version](https://cloud.google.com/logging/docs/buckets#api_3) of the instructions for creation of the linked dataset.

Maybe the documentation for the usage of the `gcloud` command needs to be fixed?

1-org/modules/centralized-logging/README.md
daniel-cit: Good catch, the option was not used by the 1-org step anymore but since the BigQuery destination is not part of the recommendation anymore we can also remove it from the auxiliary module.

1-org/modules/centralized-logging/README.md
daniel-cit: done

1-org/envs/shared/log_sinks.tf
```diff
@@ -18,15 +18,16 @@ locals {
...
+  logs_filter          = <<EOF
     logName: /logs/cloudaudit.googleapis.com%2Factivity OR
     logName: /logs/cloudaudit.googleapis.com%2Fsystem_event OR
     logName: /logs/cloudaudit.googleapis.com%2Fdata_access OR
+    logName: /logs/cloudaudit.googleapis.com%2Faccess_transparency OR
+    logName: /logs/cloudaudit.googleapis.com%2Fpolicy OR
     logName: /logs/compute.googleapis.com%2Fvpc_flows OR
     logName: /logs/compute.googleapis.com%2Ffirewall OR
-    logName: /logs/cloudaudit.googleapis.com%2Faccess_transparency
+    logName: /logs/dns.googleapis.com%2Fdns_queries
```
prateek2408, line 29: as part of this migration why is cloudauit being changed to dns_queries

1-org/envs/shared/log_sinks.tf
```diff
@@ -82,14 +72,16 @@ module "logs_export" {
     Send logs to Logbucket
   *****************************************/
   logbucket_options = {
-    logging_sink_name   = "sk-c-logging-logbkt"
-    logging_sink_filter = local.all_logs_filter
-    name                = "logbkt-org-logs-${random_string.suffix.result}"
-    location            = local.default_region
+    logging_sink_name          = "sk-c-logging-logbkt"
```
prateek2408, line 75: why is the sink name hardcoded.

1-org/modules/centralized-logging/README.md
```diff
@@ -72,8 +59,7 @@ module "logging_logbucket" {
...
```
prateek2408, line 62: are the params coming from the log analytics API doc ?

test/integration/org/org_test.go
```diff
@@ -216,58 +210,55 @@ func TestOrg(t *testing.T) {
...
 				logSink := gcloud.Runf(t, "logging sinks describe %s --folder %s", sink.name, parentFolder)
 				assert.True(logSink.Get("includeChildren").Bool(), fmt.Sprintf("sink %s should include children", sink.name))
 				assert.Equal(sink.destination, logSink.Get("destination").String(), fmt.Sprintf("sink %s should have destination %s", sink.name, sink.destination))
-				if sink.hasFilter {
-					for _, filter := range mainLogsFilter {
-						assert.Contains(logSink.Get("filter").String(), filter, fmt.Sprintf("sink %s should include filter %s", sink.name, filter))
-					}
-				} else {
-					assert.Equal("", logSink.Get("filter").String(), fmt.Sprintf("sink %s should not have a filter", sink.name))
+				for _, filter := range logsFilter {
```
prateek2408, line 258: should we check for `filter` key  `logsFilter` map first before interating over it?

prateek2408 (commented):
Added some comments.

1-org/envs/shared/log_sinks.tf
daniel-cit: actually the original entry `logName: /logs/cloudaudit.googleapis.com%2Faccess_transparency` was not replaced but moved up so that all `cloudaudit.googleapis.com` stay together in the start of the filter:

```hcl
    logName: /logs/cloudaudit.googleapis.com%2Faccess_transparency OR <=here
    logName: /logs/cloudaudit.googleapis.com%2Fpolicy OR
```

the `logName: /logs/dns.googleapis.com%2Fdns_queries` is a new entry to send DNS queries to the sink.

1-org/modules/centralized-logging/README.md
daniel-cit: the description and the parameters originate from the logbucket module
https://github.com/terraform-google-modules/terraform-google-log-export/tree/v7.7.0/modules/logbucket#inputs

test/integration/org/org_test.go
daniel-cit: we don't need to do a check for the `filter` value because the `logsFilter` list is statically defined in the code, it is not empty and all values are valid values:

```
logsFilter := []string{
	"logName: /logs/cloudaudit.googleapis.com%2Factivity",
	"logName: /logs/cloudaudit.googleapis.com%2Fsystem_event",
	"logName: /logs/cloudaudit.googleapis.com%2Fdata_access",
	"logName: /logs/cloudaudit.googleapis.com%2Faccess_transparency",
	"logName: /logs/cloudaudit.googleapis.com%2Fpolicy",
	"logName: /logs/compute.googleapis.com%2Fvpc_flows",
	"logName: /logs/compute.googleapis.com%2Ffirewall",
	"logName: /logs/cloudaudit.googleapis.com%2Faccess_transparency",
	"logName: /logs/dns.googleapis.com%2Fdns_queries",
}
```

1-org/envs/shared/log_sinks.tf
daniel-cit: The terraform example foundation is an example of how to compose the modules in  [terraform-google-modules](https://github.com/terraform-google-modules) to create a secure foundation.

Most of the Terraform resources, projects, buckets, sinks, ..., created have hardcoded names that comply with the naming conventions defined in the documentation. The naming of the resources is also part of the example.

Some resources have a random suffix added if they use a global namespace like projects.

jasonbisson (approved):
The only concern I had the change would trigger a removal of the BQ dataset, but I only saw the change to the log filter.

daniel-cit:
> The only concern I had the change would trigger a removal of the BQ dataset, but I only saw the change to the log filter.

@jasonbisson 
the new recommendation is to use log analytics with the log Bucket creating a Big query dataset associated with the log bucket.

This change removed the direct export to a BigQuery dataset that existed previously from the main code and from the inner module we use.

If this new version would be applied over a deployed version of the `master` branch it would remove an existing BigQuery dataset.

This is the reason for the `feat!:` to indicate a braking change in the code.

There is no in-place upgrade path from the a version of the foundation to a new version .

gtsorbo (approved):
LGTM

daniel-cit:
@gtsorbo @apeabody @bharathkkb build is green :)

Issue #936, opened 2023-03-01
documentation: A mandatory variable perimeter_additional_members is commented out


eeaton:
### TL;DR

Confusing documetnation and deployment steps: a mandatory variable perimeter_additional_members is commented out. 

- https://github.com/terraform-google-modules/terraform-example-foundation/blob/a4967445b4e4c3074d4dcd20dc8297813c3f2ece/3-networks-dual-svpc/common.auto.example.tfvars
- https://github.com/terraform-google-modules/terraform-example-foundation/blob/a4967445b4e4c3074d4dcd20dc8297813c3f2ece/3-networks-hub-and-spoke/common.auto.example.tfvars

In other stages, the optional variables are commented out and mandatory variables are not. 

There is a reference in [readme text](https://github.com/terraform-google-modules/terraform-example-foundation/tree/a4967445b4e4c3074d4dcd20dc8297813c3f2ece/3-networks-hub-and-spoke#deploying-with-cloud-build) that the variable needs to be manually replaced, but this can be easily overlooked because it's buried in directions

### Expected behavior

Variables that must be manually rewritten should be clearly called out both in text and in tfvars files, otherwise the developer will attempt to deploy before setting all variables and hit an error.

Tech editing the directions for usability (including the awareness that developers tend to skim repetitive steps, especially when the step contains a copy-pastable code snippet) would reduce friction for a developer to deploy.

### Observed behavior

Variable that must be manually rewritten is easily overlooked in both text and tfvars files. 

### Terraform Configuration

```hcl
n/a
```

### Terraform Version

```sh
n/a
```

obriensystems:
fmichaelobrien to review - currently on 1-org before the 2 3-networks-*

PR #1038, opened 2023-11-24
chore: update perimeter_additional_members instructions

romanini-ciandt:
This PR implements an update on `perimeter_additional_members` variable declaration in both 3-networks steps.

Basically, we want to avoid a situation where the user forget to uncomment a .tfvars line and don't have access to a project protected by the VPC service controls. With this change, if the user forgets to fulfil the variable, an error will be raised during deploy process.

### DIFF
--- a/3-networks-dual-svpc/README.md
+++ b/3-networks-dual-svpc/README.md
@@ -189,6 +189,7 @@ Run `terraform output cloudbuild_project_id` in the `0-bootstrap` folder to get 
 
    sed -i "s/REMOTE_STATE_BUCKET/${backend_bucket}/" ./common.auto.tfvars
    ```
+   **Note:** Make sure that you update the `perimeter_additional_members` variable with your e-mail in order to be able to view/access resources in the project protected by the VPC service controls.
 
 1. Commit changes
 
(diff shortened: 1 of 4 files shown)


rjerrems (changes requested):
Rather than providing an uncommented bogus parameter, why don't we make it so that `perimeter_additional_members` has no default value, so that it errors out when not provided?

In that case, if a user does not want to provide any members they can add `[""]`

romanini-ciandt:
@rjerrems thank for your inputs!
In order to solve the https://github.com/terraform-google-modules/terraform-example-foundation/issues/936 issue, I'm implementing a dummy value for `perimeter_additional_members` variable on `.tfvars` that should be replaced like we have for others required variables (i.e: in the same file we have `example.com.` for `domain`)

Nowadays, `perimeter_additional_members` variable doesn't have a default in `variables.tf`, so the current behavior is to have an error when not provided. The goal of this PR is to standardize the required variable's declaration in `.tfvars` and make all of them (the required ones) uncommented.

With this context, do you think that makes sense to keep this PR as is?

rjerrems (approved):
Ok thanks for the context - probably worth checking with @eeaton before merging to make sure it addresses the issue fully.

eeaton:
This works for me. At a future date I'm planning on doing a more thorough review of the directions in README files for this type of usability challenge, but for the scope of issue #936 that I raised, this works.

daniel-cit:
@gtsorbo this pull request edits `README.md` markdown files and `common.auto.example.tfvars` files that are not used in the integration build process.

PR #1040, opened 2023-11-29
feat: Add support to proxy-only subnetworks and new IP CIDR allocation

Samir-Cit:
Hello folks.

This PR will:
* Add support to proxy-only subnetworks in all environments and also to hub-and-spoke.
* Allocate new IP to the already existing subnetworks.
* Allocate new IP CIDR ranges to the already existing subnetworks.

```
Issue1: The IP for PSC endpoints used in code conflict with the primary subnet ranges, the IP used by PSC should be unallocated for additional regions. We need to identify an IP that doesn't break the logic of IP allocation in the document.

Issue2: The doc shows the IP Ranges with CIDR /18 and on the code is using /21, these should be consistent

Issue3: PSA requires a /16 range for each vpc, this is unreasonable. Recommendation is /20.

Issue4: PSC, PSA also need to be provisioned for common network in hub-and-spoke

Issue5: No proxy-only subnets for load balancers are provisioned. Recommended /23 size, In all subnets (including common)
```


3-networks-dual-svpc/envs/non-production/main.tf
```diff
@@ -22,40 +22,48 @@ locals {
...
   restricted_subnet_secondary_ranges = {
     (local.default_region1) = [
       {
         range_name    = "rn-${local.environment_code}-shared-restricted-${local.default_region1}-gke-pod"
-        ip_cidr_range = "100.72.128.0/21"
+        ip_cidr_range = "100.72.128.0/18"
       },
       {
         range_name    = "rn-${local.environment_code}-shared-restricted-${local.default_region1}-gke-svc"
-        ip_cidr_range = "100.72.136.0/21"
```
jasonbisson, line 58: Is there a reason were changing the subnet here instead of changing the mask to /18?

Samir-Cit: Hello @jasonbisson 
Some of the _subnets, secondary range, private service and PSC endpoint_ **IPs** also needed to be change to improve consistency between doc and code.
We had a meeting with @ellioteaton where he presented the new IPs allocation.

gtsorbo (approved):
LGTM

daniel-cit:
@gtsorbo @apeabody @bharathkkb @rjerrems 
Approve PR is green 🎉

Issue #984, opened 2023-07-02
0-org Errors in step 10 on Mac


jtblin:
### TL;DR

`find -name` fails on Mac

`sed -i` fails on Mac.

### find

On Mac, the `find` commands can be fixed by adding the `path` before `-name` e.g. 
```
for i in `find . -name 'backend.tf'`; do sed -i "s/UPDATE_ME/${backend_bucket}/" $i; done
```

### sed

The `sed` issue can be fixed by using `gnu-sed` instead: `brew install gnu-sed` and  use as `gsed SED_COMMAND` or adds to alias `alias sed=gsed` or replace `sed` by `gsed` in commands.

### Expected behavior

Command executes properly

### Observed behavior

```
find: illegal option -- n
usage: find [-H | -L | -P] [-EXdsx] [-f path] path ... [expression]
       find [-H | -L | -P] [-EXdsx] -f path [path ...] [expression]
```

### Terraform Configuration

```hcl
NA
```

### Terraform Version

```sh
Terraform v1.5.2
on darwin_arm64
```

Issue #922, opened 2023-01-17
sed on mac requires backup file 


jasonbisson:
### TL;DR

Can we add this additional comment to the document to address the requirement for Mac to have backup file for sed commands? For example, this command gives an empty argument to allow the command to complete.
 
sed -i  ''  "s/REMOTE_STATE_BUCKET/${remote_state_bucket}/" common.auto.tfvars

PR #1043, opened 2023-11-30
feat: make sed and find commands portable between Linux (GNU) and Mac OS (BSD)

romanini-ciandt:
Issues https://github.com/terraform-google-modules/terraform-example-foundation/issues/984, https://github.com/terraform-google-modules/terraform-example-foundation/issues/922

The goal of this PR is to make all the `sed` and `find` commands portable between Linux (GNU) and Mac OS (BSD). Nowadays, Mac OS users have some issues with some params (more details at the linked issues).

- For `find`: the issue is that GNU by default assumes that we want to run in the current path (or `.`), while in BSD we need to explicit pass the path (`.`) as input.

- For `sed -i`: sed file in-place suggests a need for a backup file. In GNU, if the backup file is not explict provided, the original file is overwritten but for GNU the way to achieve this same behavior (no backup saved) is to provide a zero-length extension as parameter.

References:

- For `sed`:
    - [GNU doc](https://www.gnu.org/software/sed/manual/html_node/Command_002dLine-Options.html#Command_002dLine-Options:%7E:text=If%20no%20extension%20is%20supplied%2C%20the%20original%20file%20is%20overwritten%20without%20making%20a%20backup.)
    - [BSD doc](https://man.openbsd.org/sed.1#:%7E:text=If%20a%20zero%20length%20extension,ranges%20are%20reset%20between%20files)
    - [Stack overflow discussion](https://stackoverflow.com/questions/4247068/sed-command-with-i-option-failing-on-mac-but-works-on-linux) 

- For `find`:
    - [GNU doc](https://www.gnu.org/software/findutils/manual/html_mono/find.html#Base-Name-Patterns)
    - [BSD doc](https://man.openbsd.org/find)


lanre-OG (approved):
LGTM

rjerrems (approved):

Samir-Cit:
@gtsorbo @apeabody build is green :)

daniel-cit:
@apeabody @gtsorbo Build is 🟢

PR #1064, opened 2023-12-20
chore: fix broken lint

bharathkkb:
We don't need swapper anyway since this is not a published module. I will look into either relaxing swapper constraints to exit without err or add a flag for these kind of repos. cc @daniel-cit @apeabody


daniel-cit (approved):
LGTM

PR #1065, opened 2023-12-29
fix: alternative deployment methods minor issues fix

romanini-ciandt:
**PR Overview**
The goal of this PR is to avoid an `Argument list too long` error on GitHub alternative deploy method when the bash output is too large for github actions to handle.

**When the error happens?**
The error is happening in the GH action that is triggered when the PR between plan branch and production branch is opened in 1-org step and the output is too large (too many characters).

**Which is the fix implemented?**
The fix uses a file to handle the large output and truncate the exhibition in order to limit the maximum characters. Also, it provides a hyperlink to the full action output in case the user wants to check.

In additional to that, a coupe of simple minor adjusts were made. First for a missing permission to SAs, second to correct the place of the providers in the file, third was an index being added to a Terraform Cloud resource usage and fourth to add a missing local in 1-org step.

References:
https://github.com/actions/github-script/issues/266


jasonbisson (approved):
LGTM

daniel-cit:
@apeabody @bharathkkb @gtsorbo Build is Green 🎉

bharathkkb (approved):

PR #1117, opened 2024-02-14
Draft: feat: add support to billing account logs in the organization-level

renato-rudnicki:
This pull request allows the user to logging changes on the Billing Account by the existing organization-level log sinks


1-org/modules/centralized-logging/variables.tf
```diff
@@ -25,15 +25,27 @@ variable "resources" {
...
 
   validation {
-    condition     = contains(["project", "folder", "organization"], var.resource_type)
-    error_message = "The resource_type value must be: project, organization, or folder."
+    condition     = contains(["project", "folder", "organization", "billing_account"], var.resource_type)
+    error_message = "The resource_type value must be: project, organization, billing_account or folder."
   }
 }
 
+variable "billing_account" {
```
romanini-ciandt, line 37: You need to pass an input to this variable in the module caller.

In order to do that, you should add a `billing_account = local.billing_account` input in `logs_export` module located in `/1-org/envs/shared/log_sinks.tf`.
Also, since `billing_account` will always be defined in 0-bootstrap, I don't think we should have a `default = ""` here. I would recommend no default. (If we had no default here, we would notice this missing input before due to an error)

1-org/modules/centralized-logging/variables.tf
```diff
@@ -25,15 +25,27 @@ variable "resources" {
...
+  description = "Billing Account ID used in case sinks are under billing account level. Format 000000-000000-000000."
+  type        = string
+  default     = ""
+}
+
+variable "enable_billing_account_sink" {
+  description = "Verify if billing account will be logged in the sink."
```
romanini-ciandt, line 44:
```suggestion
  description = "If true, a log router sink will be created for the billing account."
```

1-org/modules/centralized-logging/variables.tf
```diff
@@ -25,15 +25,27 @@ variable "resources" {
...
 
 variable "resource_type" {
-  description = "Resource type of the resource that will export logs to destination. Must be: project, organization, or folder."
+  description = "Resource type of the resource that will export logs to destination. Must be: project, organization, billing_account or folder."
   type        = string
 
   validation {
-    condition     = contains(["project", "folder", "organization"], var.resource_type)
-    error_message = "The resource_type value must be: project, organization, or folder."
+    condition     = contains(["project", "folder", "organization", "billing_account"], var.resource_type)
```
romanini-ciandt, line 32: I think we should remove `billing_account` from here, since we are enabling/disabling the billing account sink though `enable_billing_account_sink` variable.
If the user calls our `centralized-logging` module passing `resource_type = "billing_account"` an error will be raised due to a non-match with `parent_resource_id` (it won't be the billing account id, but the folder or org id).

romanini-ciandt (commented):
LGTM - Just left some minor suggestions

1-org/modules/centralized-logging/variables.tf
renato-rudnicki: Done.

PR #1132, opened 2024-03-03
fix(cai): modules must not define providers

oikarinen:
As per Terraform [documentation](https://developer.hashicorp.com/terraform/language/modules/develop/providers), provider configurations can be defined only in a root module.


eeaton (approved):
Thanks for the PR. 

This module was originally deployed from a separate repo, we added it in v4 so as to deploy as part of the process for 1-org stage. It seems we missed a few pieces to modify as part of making it smoothly part of this repo.

 - convention of not using a provider in a module (fixed by this PR)
 - unnecessary steps to specify a service account, should already be configured in the pipeline like other resources (fixed by this PR)
 - strange to configure CMEK only for these resources, inconsistent with how CMEK is handled everywhere else.  (I'll make a separate issue to track this)

PR #1148, opened 2024-03-11
feat!: Create project destination and remove log bucket destination for the centralized logging

daniel-cit:
This Pull request creates a new project destination for the centralized logging that replaces the existing LogBucket destination

The new project destination includes the creation of a log bucket to aggregate all the exported logs.


daniel-cit:
@apeabody @eeaton @gtsorbo 
Could you PTAL?

1-org/modules/centralized-logging/versions.tf
```diff
@@ -16,4 +16,11 @@
 
 terraform {
   required_version = ">= 1.3"
+
+  required_providers {
```
apeabody, line 20: Do we want to add this provider?  Doesn't appear be a Google or HashiCorp supported.

daniel-cit: Usage of this was a recommendation from @bharathkkb to call the Google Cloud Logging  API directly instead of using the gcloud module.

This module is presented in an official post in the Hashicorp site blog as the suggestion to make managed and unmanaged API calls for resources not supported by a provider.

https://www.hashicorp.com/blog/writing-terraform-for-unsupported-resources

daniel-cit: it is also used in the cloud run module 
https://github.com/GoogleCloudPlatform/terraform-google-cloud-run/blob/main/modules/job-exec/main.tf#L99

apeabody: Thanks @daniel-cit - Once @gtsorbo or @eeaton review, I'll get this merged.

1-org/modules/centralized-logging/README.md
```diff
@@ -61,9 +61,9 @@ module "logging_logbucket" {
...
...ket with Log Analytics enabled can be queried in the Log Analytics page using SQL queries. Cannot be disabled once enabled.<br>- retention\_days: The number of days data should be retained for the \_Default log bucket. Default 30.<br>- linked\_dataset\_id: The ID of the linked BigQuery dataset for the \_Default log bucket. A valid link dataset ID must only have alphanumeric characters and underscores within it and have up to 100 characters.<br>- linked\_dataset\_description: A use-friendly description of the linked BigQuery dataset for the \_Default log bucket. The maximum length of the description is 8000 characters. | <pre>object({<br>    logging_sink_name          = optional(string, null)<br>    logging_sink_filter        = optional(string, "")<br>    log_bucket_id              = optional(string, null)<br>    log_bucket_description     = optional(string, null)<br>    location                   = optional(string, "global")<br>    enable_analytics           = optional(bool, true)<br>    retention_days             = optional(number, 30)<br>    linked_dataset_id          = optional(string, null)<br>    linked_dataset_description = optional(string, null)<br>  })</pre> | `null` | no |
```
eeaton, line 66: Is this correct?
logging_sink_filter [...] 'Default is "" which exports all logs"

I thought the filter would be set to the locals block logs_filter in 1-org/envs/shared/log_sinks.tf

1-org/modules/centralized-logging/main.tf
```diff
@@ -118,52 +117,115 @@ resource "time_sleep" "wait_sa_iam_membership" {
...
+  headers = {
+    Authorization = "Bearer ${data.google_client_config.default.access_token}"
+    Content-Type  = "application/json",
+  }
+  request_body = <<EOF
+{
+  "exclusions": [
+    {
+      "name": "exclude_external_logs",
+      "filter": "-logName : \"/${var.logging_destination_project_id}/\""
```
eeaton, line 204: I'm not certain if I understand the outcome of this code, so allow me to double-check...

1. There is an aggregated log sink to send audit logs from all projects to `logging_destination_project_id`.
2. There is an additional log sink to send all logs from that project sink to the `aggregatedLogs` bucket sink
3. The exclusion removes logs that originated in `logging_destination_project_id` from #2. This prevents double storing logs from the log project in the `aggregatedLogs` bucket.

1-org/modules/centralized-logging/README.md
daniel-cit: The module is designed in a generic way compatible with the main module in the `log-export` repository, so that this module could be in the future moved to the log-export repository.

In the log export module the default filter is `""`,  the empty filter.
https://github.com/terraform-google-modules/terraform-google-log-export/blob/master/variables.tf#L22C1-L26C2

This is the same configuration we use for all the destinations in the centralized-logging module , the default filter is also `""` (empty) for the Pub/Sub and Storage destinations.

But when the module is used in `1-org/envs/shared/log_sinks.tf` the filter used is

```
  logs_filter          =
    logName: /logs/cloudaudit.googleapis.com%2Faccess_transparency OR
    logName: /logs/cloudaudit.googleapis.com%2Fpolicy OR
    logName: /logs/dns.googleapis.com%2Fdns_queries
}
```

1-org/modules/centralized-logging/main.tf
daniel-cit: > 1. There is an aggregated log sink to send audit logs from all projects to logging_destination_project_id.

the aggregate sink send logs from the root of the deploy and all its children.
if the foundation is deployed at org level the sink will route:
- the logs from the organization
- the logs from all folders
- the log form all projects

> 2. There is an additional log sink to send all logs from that project sink to the aggregatedLogs bucket sink

Correct

> 3. The exclusion removes logs that originated in logging_destination_project_id from feat: Adding org module to example #2. This prevents double storing logs from the log project in the aggregatedLogs bucket.

the exclusion is removing logs that does not contain the logging project ID in the logName from been routed to the `_Default` log bucket so that the _Default log bucket only contains logs generate in its own project(the loggong project)
the _Required log bucket already excludes log from outside of the project

eeaton: Got it, thanks for confirming.

eeaton (approved):

1-org/modules/centralized-logging/versions.tf
eeaton: I've reviewed, the functional changes are LGTM

PR #1196, opened 2024-04-24
chore(tfvars): fix broken symlinks

eeaton:
Fix broken symlinks. Instructions on github are to replace all *.example.tfvars with *.tfvars. Most symlinks for tfvars correctly point to a the *.tfvars files, so that a user following the instructions as written does not need to modify any symlinks.

A small number of files have symlinks to point to *.example.tfvars, which causes deployment errors related to undefined variables and requires troubleshooting to identify and fix. I've corrected these broken symlinks.

### DIFF
--- a/2-environments/envs/nonproduction/terraform.tfvars
+++ b/2-environments/envs/nonproduction/terraform.tfvars
@@ -1,1 +1,1 @@
-../../terraform.example.tfvars
+../../terraform.tfvars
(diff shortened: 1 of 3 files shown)


daniel-cit (approved):
LGTM

fmichaelobrien (approved):
/lgtm
Very nice - thanks for this, I am doing a 2nd clean org deploy this week and will reverify

apeabody (approved):
Thanks @eeaton! LGTM

mariammartins:
gcbrun

PR #1201, opened 2024-04-25
chore(golang): invalid Go toolchain version

eeaton:
Address security issue raised by codeQL, "invalid Go toolchain version".

>As of Go 1.21, toolchain versions [must use the 1.N.P syntax](https://go.dev/doc/toolchain#version).
1.21 in test/integration/go.mod does not match this syntax and there is no additional toolchain directive, which may cause some go commands to fail.


apeabody (approved):
Thanks @eeaton!

I believe this check is somewhat new, as I only noticed for a repo for the first time yesterday.

Issue #1231, opened 2024-05-13
fix: 1-org/envs/shared/projects.tf interconnect and dns labels need alignment to changes in #1199


fmichaelobrien:
### TL;DR

tef minor bug for label changes in https://github.com/terraform-google-modules/terraform-example-foundation/pull/1199/files#diff-d6697e7c916ba73d6ae87ff4b1ce67cabc9b9738ab31c9ba582e2a3218982838L279

https://github.com/terraform-google-modules/terraform-example-foundation/blob/master/1-org/envs/shared/projects.tf#L253
match
https://github.com/terraform-google-modules/terraform-example-foundation/blob/master/1-org/envs/shared/projects.tf#L237
```
-    application_name  = "org-dns-hub"
+    application_name  = "org-net-dns"
```

and
```
-net-interconnect
```

just like in 
https://github.com/terraform-google-modules/terraform-example-foundation/blob/master/1-org/envs/shared/projects.tf#L295

### Terraform Configuration

```hcl
cloud shell
```

### Terraform Version

```sh
1.3.10
```

### Additional information

part of upstream sync in 387
pr assigned to fmichaelobrien

fmichaelobrien:
```
fmichaelobrien-macbookpro:terraform-example-foundation fmichaelobrien$ git status
On branch gh1231-1-org-labels
Your branch is up to date with 'origin/gh1231-1-org-labels'.

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
	modified:   1-org/envs/shared/projects.tf

no changes added to commit (use "git add" and/or "git commit -a")
fmichaelobrien-macbookpro:terraform-example-foundation fmichaelobrien$ git diff
diff --git a/1-org/envs/shared/projects.tf b/1-org/envs/shared/projects.tf
index 681d857..4ba625e 100644
--- a/1-org/envs/shared/projects.tf
+++ b/1-org/envs/shared/projects.tf
@@ -176,7 +176,7 @@ module "interconnect" {
 
   labels = {
     environment       = "network"
-    application_name  = "org-interconnect"
+    application_name  = "org-net-interconnect"
     billing_code      = "1234"
     primary_contact   = "example1"
     secondary_contact = "example2"
@@ -250,7 +250,7 @@ module "dns_hub" {
 
   labels = {
     environment       = "network"
-    application_name  = "org-dns-hub"
[...]

PR #1199, opened 2024-04-25
chore(projects): fix network envcodes

eeaton:
This PR addresses #1198 .
Networking projects under the Networking folder do not follow the naming conventions or environment code labels as intended, this PR cleans up the implemented projected IDs to match the intended naming convention.

PR #1232, opened 2024-05-13
fix: #1231 - align 1-org labels

fmichaelobrien:
As part of #1231
discovered as part of 387 tef upstream sync

### DIFF
--- a/1-org/envs/shared/projects.tf
+++ b/1-org/envs/shared/projects.tf
@@ -176,7 +176,7 @@ module "interconnect" {
 
   labels = {
     environment       = "network"
-    application_name  = "org-interconnect"
+    application_name  = "org-net-interconnect"
     billing_code      = "1234"
     primary_contact   = "example1"
     secondary_contact = "example2"
@@ -250,7 +250,7 @@ module "dns_hub" {
 
   labels = {
     environment       = "network"
-    application_name  = "org-dns-hub"
+    application_name  = "org-net-dns"
     billing_code      = "1234"
     primary_contact   = "example1"
     secondary_contact = "example2"


fmichaelobrien:
Possibly merge after the following and sync this branch
https://github.com/terraform-google-modules/terraform-example-foundation/pull/1241#pullrequestreview-2064164125

eeaton (commented):
I want to make sure I understand the intent of this PR... Are you saying the value of `application_name` should match the environment code? 

Currently there is not a consistent convention for `application-name` to match the environment code. A few projects in this stage coincidentally do include the "net" environment code, but I don't think we should try to consistently include the `env` code in the application name. My reasoning:

1. For downstream use cases where I want to query project labels for all things in a given environment code, it's more reliable to query on the `env` label instead of regex matching the `application_name` label. 
2. In later stages where a given application is deployed across each of d/n/p environments, the same `application_name` is used across all environments. ([example](https://github.com/terraform-google-modules/terraform-example-foundation/blob/master/4-projects/modules/single_project/main.tf)). This is useful if I wanted to query all the projects related to a given app across multiple environments.

fmichaelobrien:
Thank you for the review eeaton, I ran into this minor change when merging upstream into a downstream branch.  I'll do another review on the changes that were done in this yaml be eod.  I am quoting from the issue #1231 below - essentially this change aligns one 2 of the changes to label naming that were done previously.
I will see if there is something I missed in the alignment and get back to the team.



Quote
tef minor bug for label changes in https://github.com/terraform-google-modules/terraform-example-foundation/pull/1199/files#diff-d6697e7c916ba73d6ae87ff4b1ce67cabc9b9738ab31c9ba582e2a3218982838L279

https://github.com/terraform-google-modules/terraform-example-foundation/blob/master/1-org/envs/shared/projects.tf#L253
match
https://github.com/terraform-google-modules/terraform-example-foundation/blob/master/1-org/envs/shared/projects.tf#L237
```
-    application_name  = "org-dns-hub"
+    application_name  = "org-net-dns"
```

and
```
-net-interconnect
```

just like in 
https://github.com/terraform-google-modules/terraform-example-foundation/blob/master/1-org/envs/shared/projects.tf#L295

eeaton:
My rationale is that the `application_name` doesn't really matter, so long as it is intelligible to the end-user (and customers will likely tweak this in their own implementation).

The changes you proposed match some arbitrary substrings between `application_name` and `env`, but it's not a consistent convention. Most other projects in 1-org/envs/shared/projects.tf do not use the `env` substring at all in `application_name`, so the changes just to interconnect and DNS projects actually introduce more inconsistencies.

I'll close this for now but feel free to re-open if you disagree.

fmichaelobrien:
Sounds good, understood, yes I would like to keep this file consistent with the naming convention, thanks eeaton for the review

PR #1241, opened 2024-05-16
chore(CI): remove business_unit_2 to reduce CI scope

eeaton:
Deleted subfolder for business_unit_2 under under step 4-projects, with updated directions in Readme for users to optionally recreate the this folder and automatically replace strings that change between bu1 and bu2.

This should help improve the flakiness of CI tests where a small chance of flaky error becomes cumulatively higher when creating a large number of projects (this reduces the test scope by 15 projects).


daniel-cit (changes requested):
Additional points that need to be edited

- https://github.com/terraform-google-modules/terraform-example-foundation/blob/master/helpers/foundation-deployer/stages/apply.go#L362C3-L362C65
- https://github.com/terraform-google-modules/terraform-example-foundation/blob/master/helpers/foundation-deployer/stages/destroy.go#L138C3-L138C65
- https://github.com/terraform-google-modules/terraform-example-foundation/blob/master/test/integration/projects-shared/projects_shared_test.go#L63C1-L67C5
- https://github.com/terraform-google-modules/terraform-example-foundation/blob/master/test/integration/projects/projects_test.go#L101C1-L121C5

eeaton:
Thanks for the feedback @daniel-cit . I removed those instances of business_unit_2 in the .go tests, and found a few similar places in the helper script to remove. I decided that leaving it in the README files is ok to illustrate what the complete structure looks like if someone does take the optional steps to duplicate bu2 folders.

fmichaelobrien (commented):
reviewing over the next 24h...

1-org/envs/shared/projects.tf
```diff
@@ -168,7 +168,7 @@ module "interconnect" {
   random_project_id        = true
   random_project_id_length = 4
   default_service_account  = "deprivilege"
-  name                     = "${local.project_prefix}-c-interconnect"
+  name                     = "${local.project_prefix}-net-interconnect"
```
fmichaelobrien, line 171: I understand a large portion of these label changes are already in main - I noticed them when I was doing an upstream sync last sunday
specific example
net-interconnect for 171
https://github.com/terraform-google-modules/terraform-example-foundation/blob/master/1-org/envs/shared/projects.tf#L171

also minor and not really functionality affecting but there is a pending PR on 2 changes that look to have been missed in the last PR on this yaml
https://github.com/terraform-google-modules/terraform-example-foundation/pull/1232/files
for #1231

fmichaelobrien (approved):

1-org/envs/shared/projects.tf
fmichaelobrien: I think we are good - I was using a 3hr old diff view - 1-org is out of the patch now
/lgtm

daniel-cit (approved):
LGTM

apeabody (approved):
Already has two approvals - lgtm

PR #1244, opened 2024-05-21
fix: removed duplicate folder in README segment 5. app-infra

mollyporph:
The readme segment 5. app-infra has a duplicate `fldr-development-bu2` folder in the output project tree


eeaton (approved):
Thanks for the PR!

PR #1251, opened 2024-05-23
chore(deps): consolidate renovate bot dependency fixes

eeaton:
Combine the outstanding PR from renovate[bot] into a single PR to reduce toil for release shepherding

- https://github.com/terraform-google-modules/terraform-example-foundation/pull/1247
- https://github.com/terraform-google-modules/terraform-example-foundation/pull/1246
- https://github.com/terraform-google-modules/terraform-example-foundation/pull/1242
- https://github.com/terraform-google-modules/terraform-example-foundation/pull/1193
- https://github.com/terraform-google-modules/terraform-example-foundation/pull/1126
- https://github.com/terraform-google-modules/terraform-example-foundation/pull/1123


apeabody (approved):
LGTM - The only downside is if there is an issue it can be more difficult to pinpoint which update is the source.

daniel-cit:
@apeabody @eeaton 
integration test need to be updated here

https://github.com/terraform-google-modules/terraform-example-foundation/blob/f22eb6682512468c3da638e1b9a03c5d74f4c2a4/test/integration/networks/networks_test.go#L291C1-L293C6
```
			"from": map[string]interface{}{
				"identity_type": "ANY_IDENTITY",
			},
```
to add the `sources` in the `egressPolicies` like is is used in the `ingressPolicies` 

```
			"from": map[string]interface{}{
				"sources": map[string][]string{
					"access_levels": {"*"},
				},
				"identity_type": "ANY_IDENTITY",
			},
```

to fix

```
step #16 - "converge-networks":   on .terraform/modules/base_env.restricted_shared_vpc.regular_service_perimeter/modules/regular_service_perimeter/main.tf line 87, in resource "google_access_context_manager_service_perimeter" "regular_service_perimeter":
Step #16 - "converge-networks":   87:           source_restriction = egress_policies.value["from"]["sources"] != null ? "SOURCE_RESTRICTION_ENABLED" : null
Step #16 - "converge-networks":     ‚ egress_policies.value["from"] is object with 1 attribute "identity_type"
```

and usage of `sources` should be fixed in https://github.com/terraform-google-modules/terraform-google-vpc-service-controls/blob/master/modules/regular_service_perimeter/main.tf
so that it should not be required.

Maybe from
```
lookup(egress_policies.value["from"]["sources"], "access_levels", []) 
```
to
```
lookup(lookup(egress_policies.value["from"],"sources",{}),"access_levels", [])
```

eeaton:
Thanks for the guidance Daniel. I've made an additional commit to the tests on this PR and raised the upstream fix as well.

fmichaelobrien:
thanks team for the update, I'll add this consolidated module update to a scheduled upstream sync for monday.  I agree there may be issues with backtracking regressions - however it is also good to have an effective consolidated minor release out at once so a full clean org deploy to 5-app-infra can be done once.  The reduce toil option is appreciated.

Issue #1248, opened 2024-05-21
KMS alignment: remove standalone KMS resources from cai-monitoring.tf


eeaton:
### TL;DR

The resources for [cai-monitoring.tf](https://github.com/terraform-google-modules/terraform-example-foundation/blob/229e53c8ca7ab6dd43305ab59f63a280c00dae55/1-org/envs/shared/cai_monitoring.tf#L4) are encrypted with CMEK, but this is inconsistent with how CMEK is enforced (or not) throughout the rest of the deployment. 

### Expected behavior

While many users of the blueprint will likely want to configure and implement CMEK on their sensitive resources, it doesn't make sense to implement this only for the resources under cai-monitoring. A customer typically expects consistent management of their CMEK design across all resources.

### Observed behavior

CMEK is implemented only for one resource in isolation

### Terraform Configuration

1. remove `module "kms"`from cai-monitoring.tf
1. remove `encryption_key` arg from `module "cai_monitoring"`
1. check other files for references to this module

### Terraform Version

```sh
n/a
```

### Additional information

PR #1132 fixed a few other issues introduced by porting a standalone repo into this repo as a module. 

A future major update to this repo will add options for applying CMEK consistently across all resources, possibly adopting [KMS Autokey](https://cloud.google.com/kms/docs/kms-with-autokey).

eeaton:
Additionally, we should remove extra KMS projects under each bu folder "prj-p-bu1-kms" etc. These are redundant purposes with the environment level KMS project "prj-p-kms" and contradict guidance from the product team that KMS management projects should be split by environment.  (Further delegation to each BU is an uncommon pattern, typically platform level compliance requirements are the responsibility of a central platform team).

eeaton:
Addressed by  1271

Issue #34, opened 2020-05-12
Add organization diagram


ocervell:
Add an organization diagram showing what the org will roughly look like

sbadakhc:
I think this would be really useful. How that maps to folder structure would also be really good. I would like to compare how I'm organising my folders, projects and modules with your reference architecture. I'm trying to inherit tfvars as I'm building my cloud to try and keep things DRY. I would be happy to feedback regarding this effort.

rjerrems:
Yeah I agree this would be useful, just need to make sure to keep it up to date as the repo evolves. The text based tree hierarchy in the main README is just a stop gap until we have something better. Any body interested in contributing this?

morgante:
FYI I actually have a preliminary tool to autogenerate org diagrams using gcpdraw we might be able to leverage: https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/pull/536

sbadakhc:
Thanks for feeding back gents. Much appreciated as always.

rjerrems:
For in depth diagrams / commentary - we are now going to reference the white paper here https://services.google.com/fh/files/misc/google-cloud-security-foundations-guide.pdf

PR #1271, opened 2024-06-21
chore(KMS): cleanup isolated and redundant KMS resources

eeaton:
Address #1248: 
 - remove standalone CMEK configuration applied only to cai_monitoring module (tech debt from copying this module from somewhere else)
 - remove prj-$env-$bu-kms projects which are contradictory to best practice guidance on how to manage KMS host keys.  Use the pre-existing prj-$env-kms projects instead.


0-bootstrap/main.tf
```diff
@@ -60,7 +60,6 @@ module "seed_bootstrap" {
   project_prefix                 = var.project_prefix
   encrypt_gcs_bucket_tfstate     = true
   key_rotation_period            = "7776000s"
-  kms_prevent_destroy            = !var.bucket_tfstate_kms_force_destroy
```
daniel-cit, line 63: This three inputs work as a set: 

```hcl
module "kms" {
  count   = var.encrypt_gcs_bucket_tfstate ? 1 : 0 <======
  source  = "terraform-google-modules/kms/google"
  version = "~> 2.1"

  project_id           = module.seed_project.project_id
...
  key_rotation_period  = var.key_rotation_period <======
...
  encrypters = [
    "serviceAccount:${data.google_storage_project_service_account.gcs_account.email_address}",
  ]
  prevent_destroy = var.kms_prevent_destroy <======
}
```
https://github.com/terraform-google-modules/terraform-google-bootstrap/blob/master/main.tf#L110C1-L115C56

Removing only the `kms_prevent_destroy` will make the build fail on the destroy of  `0-bootstrap`

daniel-cit (changes requested):
Maybe regarding CMEK in `0-bootstrap`it cloud be made optional since the common KMS project does not exist yet and the user may need to enable CMEK on the Terraform state bucket

eeaton:
You're right Daniel, I was working through some incremental changes and seeing the CI results, it has produced this error when tearing down the bootstrap stage:

> Step #34 - "destroy-bootstrap":         	            	Resource module.seed_bootstrap.module.kms[0].google_kms_crypto_key.key[0] has
Step #34 - "destroy-bootstrap":         	            	lifecycle.prevent_destroy set, but the plan calls for this resource to be
Step #34 - "destroy-bootstrap":         	            	destroyed. To avoid this error and continue with the plan, either disable
Step #34 - "destroy-bootstrap":         	            	lifecycle.prevent_destroy or reduce the scope of the plan using the -target
Step #34 - "destroy-bootstrap":         	            	option.}
Step #34 - "destroy-bootstrap":         	Test:       	TestBootstrap

I'll revise this.

And regarding _"Maybe regarding CMEK in 0-bootstrapit cloud be made optional since the common KMS project does not exist yet and the user may need to enable CMEK on the Terraform state bucket"..._
Right, it was my intent to not modify the CMEK on the bootstrap state bucket yet, but I made a mistake on the last commit. I'll be recommending we convert to [KMS Autokey](https://cloud.google.com/kms/docs/kms-with-autokey) as soon as it's GA, this will simplify the code and operation significantly from the current state. However, we'll need more design work to plan how to implement that at the bootstrap stage.

0-bootstrap/main.tf
eeaton: I undid this change with commit 29e0ec8

eeaton:
All checks are green... @daniel-cit and @apeabody , can I get approval please?

daniel-cit (approved):
LGTM

sleighton2022:
LGTM

PR #1210, opened 2024-04-29
fix(VPCSC): enable dryrun mode

eeaton:
Address issue #1209 .
CI tests have high flaky failure rate due in part to propagation delays with VPCSC perimeter. Changing this to dry mode better adheres to best practices for enablement and will avoid this issue in CI tests.

PR #1294, opened 2024-07-09
fix(docs): Update 3-networks-dual-svpc README after accidental change

bdashrad:
I believe this was accidentally changed in when updating README here: https://github.com/terraform-google-modules/terraform-example-foundation/pull/1210/files#diff-103694d8c8f2d87c3ec2f998d3bee4062552b630380365062470339c52531465L61


eeaton:
Good catch, that was indeed an accidental readme change with the vpcsc dryrun pr.

eeaton (approved):

PR #1300, opened 2024-07-24
Cachiman Create go.yml

WontonSam:


daniel-cit:
Thanks for your pull request but this is not a Go project, it is mainly a Terraform project.

Closing since it is unrelated to the repository

PR #1307, opened 2024-07-30
fix: remove LOG_EXECUTION_ID from environment variables

daniel-cit:
This PR removes the `LOG_EXECUTION_ID` previously added to the  environment variables of the Cloud Function gen2 used in step `1-org`


eeaton:
@daniel-cit can I close this PR now that 1306 is merged and CI tests are running successfully?

PR #1334, opened 2024-09-04
fix: fix location variables for step 4-projects in deploy helper

daniel-cit:
This PR fixes the two location variables (KMS and GCS) for step 4-projects in the deploy helper.
Instructions about enablement of the Security Command Center  to the Helper README were also added.


eeaton (approved):

daniel-cit:
@apeabody Could you PTAL?

Issue #1166, opened 2024-03-18
Bucket name length exceeds 63 characters with non-US region deployment


mromascanu123:
### TL;DR

When deploying 1-org
Error: error: bucket name validation failed bkt-cai-monitoring-5003-sources-699933893618-northamerica-northeast2
  with module.cai_monitoring.module.cloudfunction_source_bucket.google_storage_bucket.bucket,
  on ../../../terraform-google-modules/cloud-storage/google/modules/simple_bucket/main.tf line 17, in resource "google_storage_bucket" "bucket":
  17: resource "google_storage_bucket" "bucket" {

### Expected behavior

Limit name to max 63 characters regardless of how long the region name is

### Terraform Configuration

```hcl
in 0-bootstrap terraform.tfvars
org_id = "9468...etc..." # format "000000000000"
billing_account = "01A591-... etc..." # format "000000-000000-000000"

group_org_admins = "gcp-org-admins@gcp...etc....ca"

group_billing_admins = "gcp-billing-admins@gcp...etc...ca"

group_billing_data_users  = "gcp-billing-data-users@gcp...etc...ca"
group_audit_data_users    = "gcp-audit-data-users@gcp...etc...ca"
group_monitoring_workspace_users = "gcp-monitoring-workspace-users@gcp...etc...ca"

default_region = "northamerica-northeast2"
```

### Terraform Version

```sh
[myself@linuxbox 0-bootstrap]# terraform version
Terraform v1.6.0
on linux_amd64
+ provider registry.terraform.io/hashicorp/google v5.20.0
+ provider registry.terraform.io/hashicorp/google-beta v5.20.0
+ provider registry.terraform.io/hashicorp/null v3.2.2
+ provider registry.terraform.io/hashicorp/random v3.6.0
+ provider registry.terraform.io/hashicorp/time v0.11.1

Your version of Terraform is out of date! The latest version
is 1.7.5. You can update by downloading from https://www.terraform.io/downloads.html
[myself@linuxbox 0-bootstrap]#
```

### Additional information

Very easy (and tested) fix also guaranteeing uniqueness
in 1-org/modules/cai-monitoring/main.tf
locals {
  project_service_apis = [
  ... etc ...
  ]
 cai_source_name = var.random_suffix ? "CAI Monitoring - ${random_id.suffix.hex}" : "CAI Monitoring"
  ## add this:
[...]

PR #1347, opened 2024-10-09
Fix: bucket name length exceeds 63 characters with non-US region deployment

mariammartins:
Hey folks, this PR contains changes to the structure of bucket names, so that when the deployment is made in a region with a long name, such as "northamerica-northeast1", the bucket does not exceed the 63 character limit, regarding the issue [(1166)](https://github.com/terraform-google-modules/terraform-example-foundation/issues/1166)


gtsorbo (approved):
LGTM

PR #1367, opened 2025-01-08
Fix: Fixes for Github deploy

renato-rudnicki:
This PR fix unsupported argument for `folder_deletion_protection` and fixes `attribute_condition` value in `gh_oidc` module.

### DIFF
--- a/0-bootstrap/github.tf.example
+++ b/0-bootstrap/github.tf.example
@@ -93,7 +93,7 @@ module "gh_cicd" {
 
 module "gh_oidc" {
   source  = "terraform-google-modules/github-actions-runners/google//modules/gh-oidc"
-  version = "~> 4.0.0"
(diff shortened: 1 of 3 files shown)


0-bootstrap/github.tf.example
```diff
@@ -93,12 +93,13 @@ module "gh_cicd" {
 
 module "gh_oidc" {
   source  = "terraform-google-modules/github-actions-runners/google//modules/gh-oidc"
-  version = "~> 3.1"
+  version = "~> 4.0.0"
```
apeabody, line 96: Do we need to pin to the patch?

```suggestion
  version = "~> 4.0"
```

renato-rudnicki: No, I will update. Good catch!

apeabody (approved):

PR #1398, opened 2025-04-02
Fix: Update Gitlab and GitHub steps

renato-rudnicki:
This PR update steps for deploy Github and Gitlab


eeaton:
@renato-rudnicki It looks like all the files from the recent PR to remove the base project are also included as changes in this PR...

Can you try updating to `main` one more time? I think that will reduce the scope from the "files changed" tab.

PR #1402, opened 2025-04-11
Feat: enable kms centralized key usage tracking

guilherme-hosoda-cit:
This PR enables the centralized [View Key Usage](https://cloud.google.com/kms/docs/view-key-usage) KMS feature.


1-org/envs/shared/variables.tf
```diff
@@ -26,6 +26,12 @@ variable "enable_scc_resources_in_terraform" {
   default     = false
 }
 
+variable "enable_kms_key_usage_tracking" {
+  description = "Enable KMS centralized key usage tracking system."
+  type        = bool
+  default     = false
```
eeaton, line 32: I recommend setting the default to `true` because this is a recommended feature, and it doesn't have side-effects like extra cost that customers might want to avoid.

1-org/envs/shared/iam.tf
```diff
@@ -154,3 +166,10 @@ resource "google_project_iam_member" "cai_monitoring_builder" {
   role   = each.key
   member = "serviceAccount:${google_service_account.cai_monitoring_builder[0].email}"
 }
+
+resource "google_organization_iam_member" "kms_protected_resources_viewer" {
```
eeaton, line 170: I don't think we need an extra group for this, this can be removed.

Typically, the persona at a customer with the KMS Admin roles is the persona who wants to view KMS resources across the org.  I can't think of a common scenario where a customer would want separate groups for kms_admin and kms_protected_resources_viewer

1-org/envs/shared/variables.tf
```diff
@@ -141,15 +147,17 @@ variable "gcp_groups" {
...
-    scc_admin            = optional(string, null)
-    global_secrets_admin = optional(string, null)
-    kms_admin            = optional(string, null)
+    audit_viewer                   = optional(string, null)
+    security_reviewer              = optional(string, null)
+    network_viewer                 = optional(string, null)
+    scc_admin                      = optional(string, null)
+    global_secrets_admin           = optional(string, null)
+    kms_admin                      = optional(string, null)
+    kms_protected_resources_viewer = optional(string, null)
```
eeaton, line 160: Same comment as the IAM binding, I don't think we need an extra group for kms_protected_resources_viewer. This can be removed.

1-org/README.md
```diff
@@ -82,6 +82,8 @@ to Bigquery and Pub/Sub. This will result in additional charges for those copies
 
 - To use the **hub-and-spoke** architecture described in the **Networking** section of the [Google Cloud security foundations guide](https://cloud.google.com/architecture/security-foundations/networking#hub-and-spoke), set the `enable_hub_and_spoke` variable to `true`.
 
+- As a KMS Administrator, you can [view all the keys in your organization from a centralized location](https://cloud.google.com/kms/docs/view-key-usage) by setting the `enable_kms_key_usage_tracking` variable to true.
```
eeaton, line 85: Because we will set `enable_kms_key_usage` tracking to true, I recommend to also remove the extra directions from the readme. This helps keep the instructions more simple.

The documentation on what the variable does is already auto-generated in the readme.md under envs/shared, so we don't need redundant text here to describe each variable.

eeaton (approved):
Thanks for the updates, LGTM

PR #1406, opened 2025-04-23
Chore: Use default credentials for tests instead of SA key

guilherme-hosoda-cit:
This PR introduces support for default Google Cloud SDK (gcloud) credentials during testing. This eliminates the need for a service account key.

> [!NOTE]
> This solution injects your entire Gcloud SDK into the CFT container.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -49,10 +49,10 @@ To use this setup, you need a service account with:
 - Folder Creator and Project Creator within a folder/organization.
 - Billing Account Administrator on a billing account
 
-Export the Service Account credentials to your environment like so:
+Export the Service Account email to your environment like so:
 
 ```bash
-export SERVICE_ACCOUNT_JSON=$(< credentials.json)
+export GOOGLE_IMPERSONATE_SERVICE_ACCOUNT=account-id@my-project-id.iam.gserviceaccount.com
 ```
 
 You will also need to set a few environment variables:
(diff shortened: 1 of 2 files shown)


eeaton:
Thanks Guilherme for the PR!

Unfortunately, there are more changes needed. When I tried testing this locally it fails at an error related to service account keys. (Service Account keys are blocked for most new customer accounts with [secure by default org policies](https://cloud.google.com/resource-manager/docs/secure-by-default-organizations)), and the automation in the CFT image still relies on service account keys.

I encounter this running `make docker_test_prepare` to create the resources end-to-end. the first step to create a folder like "test_foundation_folder_z7orq3" and bootstrap project like "ci-foundation-z7orq3-afkk" work, but then fails with an error like the following:

"""
_Key creation is not allowed on this service account. [google.rpc.error_details_ext] { message: \"Key creation is not allowed on this service account.\" details { [type.googleapis.com/google.rpc.PreconditionFailure] { violations { type: \"constraints/iam.disableServiceAccountKeyCreation\" subject: \"projects/ci-foundation-z7orq3-afkk/serviceAccounts/ci-account@ci-foundation-z7orq3-afkk.iam.gserviceaccount.com?configvalue=ci-account%40ci-foundation-z7orq3-afkk.iam.gserviceaccount.com\" description: \"Key creation is not allowed on this service account._
"""

In this repo, I found there is a resource google_service_account_key.int_test at test/setup/iam.tf, and `output sa_key` under outputs.tf also references that value. It looks like I can get around the issue by removing those two references. 

Can you please update this PR to also remove all references to google_service_account_key.int_test?

PR #1408, opened 2025-04-29
fix: Fix for google-terraform-provider in version 6.26.0

renato-rudnicki:
This PR removes the exception for version 6.26.0 in the google-terraform-provider. Thread for the bug (already fixed): https://github.com/hashicorp/terraform-provider-google/issues/21950


apeabody (approved):
Thanks @renato-rudnicki!

I'm comfortable with this change, but if there is a concern you could still specifically exclude those versions: `!= 6.26.0, != 6.27.0`.

PR #1459, opened 2026-02-19
fix!: Update 'location' parameter for Gen2 compatibility

cristianovi:
#### What does this PR do?
This PR updates the `function_location` parameter to `location` in the `/terraform-example-foundation/1-org/modules/cai-monitoring/main.tf` file.

#### Why is this change necessary?
This change is required to align the code with the new module version's requirements, which was updated to better support Cloud Functions Gen2.

#### Error Encountered Before This Fix

```Error: Missing required argument
on ../../modules/cai-monitoring/main.tf line 143, in module "cloud_function":
143: module "cloud_function" {
The argument "location" is required, but no definition was found.

### DIFF
--- a/1-org/modules/cai-monitoring/main.tf
+++ b/1-org/modules/cai-monitoring/main.tf
@@ -142,7 +142,7 @@ resource "google_scc_v2_organization_source" "cai_monitoring" {
 // Cloud Function
 module "cloud_function" {
   source  = "GoogleCloudPlatform/cloud-functions/google"
-  version = "~> 0.6"
+  version = "~> 0.8.0"
 
   function_name         = "caiMonitoring"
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: 1-org/modules/cai-monitoring/main.tf
Comment (guilherme-hosoda-cit) [inline 1-org/modules/cai-monitoring/main.tf]: Since no major version of the module has been released yet, I recommend pinning it to the MINOR version instead of the MAJOR version to avoid these problems in the future.

```suggestion
  version = "~> 0.8.0"
```

guilherme-hosoda-cit (changes requested):

guilherme-hosoda-cit (approved):

amandakarina (approved):

marcos-leal-cit (approved):

apeabody (approved):
Thanks @cristianovi 

Reviewed by @amandakarina

